PR #51, opened 2015-01-02
Using recent version of commander with co-prompt

palanik:
Not a convincing reason to get stuck with commander v1.3.2, just to use program.confirm().
Upgrade to more recent, stable release of commander and use @tj's [co-prompt](https://github.com/tj/co-prompt).

### DIFF
--- a/bin/express
+++ b/bin/express
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 
 var program = require('commander');
+var prompt = require('co-prompt');
 var mkdirp = require('mkdirp');
 var os = require('os');
 var fs = require('fs');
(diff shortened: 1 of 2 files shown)


dougwilson:
Nice!

bin/express
```diff
@@ -65,7 +66,7 @@ var www = fs.readFileSync(__dirname + '/../templates/js/www', 'utf-8');
     if (empty || program.force) {
       createApplicationAt(path);
     } else {
-      program.confirm('destination is not empty, continue? ', function(ok){
+      prompt.confirm('destination is not empty, continue? ')(function(err,ok){
```
rlidwka: If there is an `err`, shouldn't it be handled somehow? `if (!err && ok)` maybe? Not sure what `co-prompt` api is.

dougwilson: I just read the source code for `co-prompt` and, in the latest version, it's not possible for `err` to contain anything other than `null`, but I'm sure it was put there as a matter of a standard interface. I still think it should be handled anyway, though :) A `if (err) throw err` should be good enough for this command line program.

palanik: @rlidwka @dougwilson I agree that err should be handled. I am indirectly handling err in my code, such that it would fall thru the else part and abort.

Feel free to implement, you think is the right way. 
Cheers.

PR #52, opened 2015-01-07
Using fs.writeFileSync in place of fs.writeFile.

palanik:
No callback argument provided for asynchronous version of fs.writeFile() method used in code.
Explicitly use synchronous version - fs.writeFileSync()

### DIFF
--- a/bin/express
+++ b/bin/express
@@ -260,7 +260,7 @@ function emptyDirectory(path, fn) {
  */
 
 function write(path, str, mode) {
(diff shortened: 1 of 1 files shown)


palanik:
The node docs have been changed that the callback argument is no longer optional for [fs.writeFile()](http://nodejs.org/docs/v0.10.35/api/fs.html#fs_fs_writefile_filename_data_options_callback)

dougwilson:
Thanks!

palanik:
Thanks.

PR #53, opened 2015-01-10
Bumped EJS dependency to ~2.0.0

mde:
Work on v2 is happening here: https://github.com/mde/ejs I'm the new package maintainer.

### DIFF
--- a/bin/express
+++ b/bin/express
@@ -199,7 +199,7 @@ function createApplicationAt(path) {
         pkg.dependencies['jade'] = '~1.8.2';
         break;
       case 'ejs':
-        pkg.dependencies['ejs'] = '~1.0.0';
+        pkg.dependencies['ejs'] = '~2.0.0';
         break;
       case 'hjs':
         pkg.dependencies['hjs'] = '~0.0.6';


dougwilson:
Thanks, @mde :) This would have actually happened earlier, but when I tried, my tests failed, which is why I ended up filing those two bug reports to you :)!

mde:
Thank you so much for those fixes! I was in such a hurry to reimplement the few missing features and land v2, that I hadn't linted the code. You may notice there's now a .jshintrc file, so there shouldn't be any regressions of that type. I'll be excited to see more people moving up to v2!

dougwilson:
Yea, me too. A few months ago I was considering volunteering to take over ejs, but never did because I don't have the time to take that on :) Glad to see it with an owner once again.

mde:
:+1:

dougwilson:
I altered you PR to `~2.1.4`

mde:
Sounds great. They'll both pick up minor version updates. Thanks for merging this so swiftly.

dougwilson:
No problem :) Feel free to send over a PR if there is some time you really want the version here bumped and I haven't already done so !

PR #55, opened 2015-01-13
gen: sort dependencies

outsideris:
Dependencies in `package.json` will be sorted when a user use npm with `--save`.

So, `package.json` will be like after `npm install --save mocha` is run:

```shell
diff --git a/package.json b/package.json
index ddf6009..8127d63 100644
--- a/package.json
+++ b/package.json
@@ -6,12 +6,13 @@
     "start": "node ./bin/www"
   },
   "dependencies": {
-    "express": "~4.10.6",
     "body-parser": "~1.10.1",
     "cookie-parser": "~1.3.3",
-    "morgan": "~1.5.1",
-    "serve-favicon": "~2.2.0",
     "debug": "~2.1.1",
-    "jade": "~1.8.2"
+    "express": "~4.10.6",
+    "jade": "~1.8.2",
+    "mocha": "^2.1.0",
+    "morgan": "~1.5.1",
+    "serve-favicon": "~2.2.0"
   }
```

If generator generator the `package.json` with sorted dependencies.
it will be like after `npm install --save mocha` is run:

```shell
diff --git a/package.json b/package.json
index b2c6c18..8d6e23c 100644
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "debug": "~2.1.1",
     "express": "~4.10.6",
     "jade": "~1.8.2",
+    "mocha": "^2.1.0",
     "morgan": "~1.5.1",
     "serve-favicon": "~2.2.0"
   }
```

### DIFF
--- a/bin/express
+++ b/bin/express
@@ -2,6 +2,7 @@
 
 var program = require('commander');
 var mkdirp = require('mkdirp');
+var sortedObject = require('sorted-object');
 var os = require('os');
 var fs = require('fs');
 var path = require('path');
@@ -223,6 +224,8 @@ function createApplicationAt(path) {
         break;
       default:
     }
+
+    pkg.dependencies = sortedObject(pkg.dependencies);
 
     write(path + '/package.json', JSON.stringify(pkg, null, 2));
     write(path + '/app.js', app);
--- a/package.json
+++ b/package.json
@@ -26,7 +26,8 @@
   "license": "MIT",
   "dependencies": {
     "commander": "2.6.0",
-    "mkdirp": "0.5.0"
+    "mkdirp": "0.5.0",
+    "sorted-object": "1.0.0"
   },
   "main": "bin/express",
   "preferGlobal": true,


dougwilson:
There is no absolute guarantee that `JSON.stringify` will serialize object keys in the same order of insertion. If we what this to be a feature, we either need to write our JSON by hand (the approach `mime-db` takes) or use some module on `npm` that does it so we're not responsible for bugs from it.

outsideris:
How about [sorted-object](https://www.npmjs.com/package/sorted-object)? Actually [npm use it](https://github.com/npm/npm/blob/master/lib/install.js#L85) and I borrow my code from it.

dougwilson:
I'm :+1: on just depending on `sorted-object`, especially if that's what `npm` is going to use to sort `dependencies`, which was the reason you wanted it :) Then they should always agree on the correct sort :)

outsideris:
@dougwilson Thank you for your feedback. I didn't know you like adding dependency or not. :) I will update PR

dougwilson:
It's no problem, and definitely understandable :) In this case, I think a dep actually works out the best of all since `npm` is using it. It basically makes the PR technically become "sort `dependencies` the same as `npm(1)` does" :D

outsideris:
I added `sorted-object`. let me know if there are some problem or breaking convention.

dougwilson:
Great, it looks good to me :) This will be part of the 4.11 generator release :)

dougwilson:
Thanks. I altered your commit to add a test, so it can be considered a feature and doesn't break :)

outsideris:
Good!!

PR #56, opened 2015-01-14
add next param to the router functions

### LINKED ISSUE #57: add next param to the router function
in almost all the router functions doing db kind of operations, we want to do ``if (err) next(err);``

next is used by almost all the router functions. I think it makes sense to add that as a param in the router functions generated by this generator.

chirag04:
fixes #57


dougwilson:
Thanks

PR #58, opened 2015-01-15
PORT can be named pipe when using iisnode.

rramachand21:
Users running express generator to create apps that run on windows using iisnode have to modify this line in the bin/www file to make the apps work with iisnode because iisnode and node.exe communicate over named pipes. In the iisnode case, the process.env.PORT is a named pipe and not a tcp port. To support this, please accept this pull request.

### DIFF
--- a/templates/js/www
+++ b/templates/js/www
@@ -12,7 +12,7 @@ var http = require('http');
  * Get port from environment and store in Express.
  */
 
-var port = parseInt(process.env.PORT, 10) || 3000;
+var port = process.env.PORT || 3000;
 app.set('port', port);
 
 /**
--- a/templates/js/www
+++ b/templates/js/www
@@ -58,5 +58,5 @@ function onError(error) {
  */
 
 function onListening() {
-  debug('Listening on port ' + server.address().port);
+  debug('Listening on port ' + (server.address().port || server.address()));
 }


dougwilson:
Will do, but I need someone to answer this question first: https://github.com/strongloop/express/issues/2498#issuecomment-69700914

rramachand21:
"Can you tell me what happens on line 61 when the port is a named pipe?" -- parseInt("namedpipe") return NaN which is not a valid named pipe and hence when iisnode tries to connect to this, it fails.

dougwilson:
I understand that line, but I'm talking about line 61, not the line you changed.

dougwilson:
This is line 61, if you don't have the bin/www file handy:

    debug('Listening on port ' + server.address().port);

rramachand21:
Oh.. without my change, the line 61 prints this -- 
Listening on port 3000

with my change, the line 61 prints this - Listening on port undefined
but the app works and just before line 28, i print the port and it says :

"calling server.listen with \\.\pipe\4323f7c5"

dougwilson:
I know. Basically, what I'm getting at is that there's "this is the bare minimum change that just happens to make it work" and "I've updated all the generated files and all the features to correctly work with named pipes" is all :)

So, I haven't gotten around to test this, but since you have a way, please, if possible, update the PR to add full named pipe support to all the feature of the generated app, including things like that debug message.

Bonus points if you add a test so named pipe support doesn't break in the future.

rramachand21:
Ok, Just found out that server.address() shows the named pipe correctly. server.address().port is undefined because its a named pipe.

dougwilson:
Nice :) Just change it to whatever works for both cases :) I just don't have a good way to test named pipes right now to look into it, so I'm just relying on how you change the code here.

templates/js/www
```diff
@@ -58,5 +58,5 @@ function onError(error) {
  */
 
 function onListening() {
-  debug('Listening on port ' + server.address().port);
+  debug('Listening on port ' + (server.address().port || server.address()));
```
dougwilson: I assume this will print `Listening on port \.\pipe\4323f7c5`, is that correct?

rramachand21: yes, it prints Listening on port \.\pipe\4323f7c5 in case of iisnode and it prints the integer port if the port was set to integer.

dougwilson: Cool :) I was wondering how Node.js would have fit in the named pipe to `{host: ..., port: ...}`... just just not even giving an object, haha.

dougwilson: Can you do one last thing, pretty please? See if this code works right:
```js
var addr = server.address()
var bind = typeof addr === 'string'
  ? 'pipe ' + addr
  : 'port ' + addr.port
debug('Listening on ' + bind)
```

rramachand21: Yup, works perfect -- Dont know why the semicolon is missing at the end line..but otherwise looks good.

dougwilson: Thanks :) It was a typo and JS doesn't require semicolons in most places, so I didn't realize.

PR #60, opened 2015-01-20
fix prompt message for `debug` command

haruair:
Debug environment variable was changed. The command in prompt message need to be updated.


dougwilson:
Thank you!

PR #63, opened 2015-01-27
Renamed template .gitignore file to work around npm issue

platta:
Renamed template .gitignore file to gitignore so that npm will not exclude it during packaging (see npm/npm#1862).

Per the discussion on the npm issue, if there is no .npmignore file but there is a .gitignore file, it will be included in the npm package as .npmignore. If there is an .npmignore file and also a .gitignore file, the .gitignore file will simply be excluded from the package.

(This is my first pull request, so I apologize for any improper github etiquette or other mistakes)


dougwilson:
It's no problem. I assume what you're getting at is that the `.gitignore` template file is missing from the package currently published on npm?

platta:
Yes. There is no .gitignore file in the package, but there is an .npmignore file in its place. That's how I found my way to that issue on the npm project which appears to explain the behavior.

dougwilson:
Ah. What the heck, npm :)?

PR #69, opened 2015-03-17
added console.log to www to display the port in the terminal

mjhea0:


dougwilson:
What's wrong with using the DEBUG environment? Also tests are failing.

mjhea0:
I teach using the express generator and DEBUG confuses people. Just tired of telling people to log it to the console.

dougwilson:
Gotcha. So if we make this change, we don't want to have two things logging to the console. I would say replace the existing one and the new one needs to have a way to turn it off (say, off if in production).

dougwilson:
Closing due to lack of action on the last comment. If you can address it, let me know and I can always open the PR again or you can always make a new PR :)

PR #89, opened 2015-11-11
Add --es6 to enable ES6 support

asimpletune:
Hi! I really like this project and use it all the time. One thing that I've been doing a lot more of lately though is using ES6, but I have to perform the steps in this PR every time I start a new project. I think of express-generator as being a great way to start coding immediately when you're struck with that good idea, so I think this would help serve that mission. Thanks!


dougwilson:
Cool! Would it be possible to update the README as well?

bin/express
```diff
@@ -98,10 +99,10 @@ function createApplication(app_name, path) {
...
 
   // JavaScript
-  var app = loadTemplate('js/app.js');
-  var www = loadTemplate('js/www');
-  var index = loadTemplate('js/routes/index.js');
-  var users = loadTemplate('js/routes/users.js');
+  var app = program.es6 ? loadTemplate('es6/app.js') : loadTemplate('js/app.js');
+  var www = program.es6 ? loadTemplate('es6/www') : loadTemplate('js/www');
+  var index = program.es6 ? loadTemplate('es6/routes/index.js') : loadTemplate('js/routes/index.js');
+  var users = program.es6 ? loadTemplate('es6/routes/users.js') : loadTemplate('js/routes/users.js');
```
dougwilson: If everything is going to be loaded from a split directory, is it possible to make this more maintainable? Perhaps just decide the base template directory first and then load the files from within that instead of having a ternary on every line?

templates/es6/app.js
```diff
@@ -0,0 +1,60 @@
...
+import path from 'path';
+import favicon from 'serve-favicon';
+import logger from 'morgan';
+import cookieParser from 'cookie-parser';
+import bodyParser from 'body-parser';
+
+import routes from './routes/index';
+import users from './routes/users';
+
+let app = express();
```
dougwilson: Should this be `const`?

templates/es6/app.js
```diff
@@ -0,0 +1,60 @@
...
+app.use(bodyParser.json());
+app.use(bodyParser.urlencoded({ extended: false }));
+app.use(cookieParser());{css}
+app.use(express.static(path.join(__dirname, 'public')));
+
+app.use('/', routes);
+app.use('/users', users);
+
+// catch 404 and forward to error handler
+app.use(function(req, res, next) {
```
dougwilson: Should this be the following? (not sure if it complies, but seems like it should)
```js
app.use((req, res, next) => {
  next(Object.assign(new Error('Not Found'), {
    status: 404
  });
});
```

templates/es6/www
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env node
+
+/**
+ * Entrypoint to application for es6 support
```
dougwilson: I think missing a space.

templates/es6/www
```diff
@@ -0,0 +1,96 @@
...
+ */
+ 
+require("babel-core/register");
+
+/**
+ * Module dependencies.
+ */
+
+var app = require('../app').default;
```
dougwilson: Nothing in this files seems to be in the ES6 style. Should it be updated to fit in with the rest of the files?

test/cmd.js
```diff
@@ -120,6 +121,111 @@ describe('express(1)', function () {
...
+      var file = path.resolve(dir, 'app.js');
+      var app = require(file).default;
+      request(app)
+      .get('/')
+      .expect(200, /<title>Express<\/title>/, done);
+    });
+
+    it('should generate a 404', function (done) {
+      var file = path.resolve(dir, 'app.js');
+      var app = require(file).default;
```
dougwilson: Why can it not be `require(file)`? Seems like the ES6 app is subtly different from the others and the rest of Node.js things if it falls under a `default` property.

test/cmd.js
```diff
@@ -1,4 +1,5 @@
 
+require("babel-core/register");
```
dougwilson: We need to do something different here; this will cause the tests not to detect an accidental ES6 syntax in the ES5 apps.

dougwilson:
Also, the tests keep failing on CI because apparently babel takes a really long time to install because it's so big. I guess just increase the timeout for the es6 tests?

bin/express
asimpletune: Sure!

templates/es6/app.js
asimpletune: Yup, I'll add the changes to the other applicable places.

templates/es6/app.js
asimpletune: Yeah, that'll compile. I'll refactor everything to use fat arrow.

templates/es6/www
asimpletune: What I was seeing the other night was that `require("babel-core/register");` was working as expected for all files *outside* the app entrypoint, i.e. this file, but didn't seem to affect the actual entrypoint. I'll look into this some more though to get a better idea.

test/cmd.js
asimpletune: This was by far the biggest source of consternation. I'll look into this more deeply because you're right, you should just be able to import the app without really knowing where it came from and it should work the same regardless of it being es5 or 6.

test/cmd.js
asimpletune: Hmm I see what you mean, good suggestion. Ok, so the issue is that when the project gets generated with `--es6` the templates are changed to make use of that syntax. So in the tests then, if they're importing the files generated with `--es6` to run the tests, there's going to be a syntax error unless that require statement is present. 

I'll have to think some more about how to do this, I'm open to suggestions though. One idea would be put the ES6 test code in a separate file, which could have the require statement from above, but I don't know if that breaks from what you want the tests to look like.

test/cmd.js
asimpletune: This works and ES5 and ES6 users wouldn't have to change their code when importing `app.js` as a module. 

```js
export default app

/**
 * For ES5 compatibility with require
 */
module.exports = app;
```

asimpletune:
I addressed everything discussed so far in here. One issue that came up was mocha not liking a global leak that occurs in a library when transpiling is done for the es6 test setup. I suppress that by adding __core-js_shared__ to mocha's globals option. Here's the issue I raised about that https://github.com/babel/babel-loader/issues/152.

templates/es6/www
jordi12100: You can fix this with babel-node instead of the regular node package.

nyteshade: Nothing in bin/www can be in ES6 format. The call to babel-core/register patches the require statements and any code executing other code from this file (ala ../app) will run in full ES6 compatility

nyteshade:
@asimpletune I also want this PR to go. I do this on all my projects as well and have the same feelings about express-generator. I would like to see a follow up PR to this that does ES7, i.e. babel-preset-stage-0 and babel-plugin-transform-runtime to allow the usage of async/await in the ES2015+ code. The changes are minimal and I could provide them if you're not familiar.

asimpletune:
Hey, @nyteshade, glad you like it. I think you might be right about .bashrc. I originally excluded it, because you can include in the package.json file, but it seems it may be more clear to keep it separate from the main package file. Also I did it that way originally because all the scaffolding was already there so it was easier.

asimpletune:
Also good to know about es7. I won't add it to this PR though to keep the commit/issue history clean, but maybe after this one you or I can do that.

templates/es6/routes/index.js
```diff
@@ -0,0 +1,9 @@
+import express from 'express';
```
sheepsteak: How about: `import {Router} from 'express';` ?

templates/es6/www
sheepsteak: `babel-node` isn't supposed to be used in production either and using it here might make people do it - https://babeljs.io/docs/usage/cli/#babel-node.

How about moving the contents of `www` into a `server.js` and then `www` simply becomes:
```javascript
require('babel-core/register');
require('../server');
```
We can use ES2015 almost everywhere then. Or does that make the generating harder?

sheepsteak:
I hate to be pedantic but shouldn't all the `es6` parts be replaced with `es2015`? :neckbeard:

GGross5213:
So what is going on with this? Can I currently use the Express.js generator with and es6 option?

templates/es6/www
asimpletune: Hey, wow, I must have missed this, sorry. That's a great suggestion, it doesn't make generating harder at all. The one thing that I'm not sure about is it would cause projects generated with `express --es6` to have a slightly different directory structure. So, were we to take that route, then the base js (AKA ES5) option should look the same.

This then leads to a case where some of the simplicity is gone. Currently, when you generate a project, the only initial js file in your project's top level directory is `app.js`, which is part of the appeal. That said, putting `server.js` in there as well isn't wrong, and in fact I think it's more clear (i.e. `server.js` is where you go to make changes to the actual server and `app.js` is where you go to make changes to your app), but it would be a small departure. 

Any way, sorry for not seeing any of this. I made and pushed the changes. Nice idea!

templates/es6/routes/index.js
asimpletune: I think it's better to leave the express object still accessible, since people are going to be building other projects off of this. Don't want to be too concise :)

asimpletune:
Hi @nyteshade, I pushed the changes for supporting the `.babelrc` file, rather than relying on `package.json` for babel configuration. I think going this way makes it cleaner. Thanks for the suggestion!

ES7 support would be cool, but I don't really know a lot about it, so I'll hold off on that for now. I'd rather have one-touch https support, since that's a fairly common requirement for some of the newer in-browser JS stuff.

asimpletune:
@sheepsteak and @nyteshade you are absolutely right, pushed an update and referenced the new spec http://www.ecma-international.org/ecma-262/6.0/ thanks!

nyteshade:
Some tests seem to be failing. I would like to use this if you get it fixed

asimpletune:
Great! I'm on the road right now, but when I get a chance to pull into somewhere I'll take a look, might just be related to those npm issues that got resolved a few days ago.

asimpletune:
Okey doke, all checks pass, @nyteshade. Also, I merged the feature changes with the most recent changes you've made, @dougwilson.

nyteshade:
Who is responsible for hitting merge? @asimpletune @dougwilson

blakeembrey:
Just a comment, but if this plans to land I think it should be more explicit. This isn't ES6, this is Babel-flavoured. It will not work with TypeScript. It might not even work when ES6 support is actually landed into NodeJS core. Is it possible to explain that better before pushing people toward a specific implementation of ES6 modules?

templates/es2015/server.js
```diff
@@ -0,0 +1,100 @@
+/**
+ * Module dependencies.
+ */
+
+import app from './app';
+import {default as debugModule} from 'debug';
```
blakeembrey: Why not do `import debugModule from 'debug'`?

nyteshade:
@blakeembrey 
It's a good point. Typescript isn't ES6 either. Perhaps the switch should be --babel-es2016 rather than --es2016 to be specific.

templates/es2015/server.js
asimpletune: The default module isn't named debugModule, so you have to alias it to avoid a naming collision with debug which if I remember correctly lies elsewhere.

asimpletune:
> who is responsible for hitting merge?

@nyteshade, that would be @dougwilson

asimpletune:
@blakeembrey, I'll look into these concerns next chance I get, still on the road.

templates/es2015/server.js
blakeembrey: I don't see any `debug` in this file.

blakeembrey: Also, the default module has no correlation with the import name - you could call it whatever you want, including my comment above.

blakeembrey: Here's an example, if you open both these links you'll see the generated code is identical.

* https://babeljs.io/repl/#?evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&code=import%20foo%20from%20'foo'
* https://babeljs.io/repl/#?evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&code=import%20%7Bdefault%20as%20foo%7D%20from%20'foo'

dougwilson:

asimpletune:
No worries, I'm in glacier right now so almost no connection, but I'll reopen a PR momentarily and it will address all of the suggestions from this thread.

AlJohri:
hi @dougwilson, do you know what's the status on this PR? I see that #123 is also closed

asimpletune:
@AlJohri hi, I saw this comment and merged the most recent changes in the project with the ES2015 code in #146

PR #92, opened 2015-12-12
Add 1 template option instead of 3

jordi12100:
Having different options can cause confusion, that’s why I think it’s best to get rid the separate template options.

If we continue having separate params, then we may end up having 10 different options.
It’s also better to avoid -H, because -h is mostly used by displaying help information.

As an addition, adding both --hogan and --ejs doesn't add both and thus it shouldn't be presented like they add both.


jordi12100:
Also updated the readme file to match with the new command

dougwilson:
Thanks! Seems like a good idea. With the removal of existing options, it would make it non-backwards-compatible. We should probably wait until the next major, and since this version tracks Express's version, we'll wait until Express 5 is ready :) !

dougwilson:

PR #95, opened 2015-12-13
Added tests for less engine

jordi12100:
Other engines look similar. This is a good start.

### DIFF
--- a/test/cmd.js
+++ b/test/cmd.js
@@ -151,7 +151,7 @@ describe('express(1)', function () {
       assert.notEqual(files.indexOf('package.json'), -1, 'should have package.json file');
     });
 
-    it('should replace css tag with right engine', function (done) {
+    it('should replace css tag with less engine', function (done) {
       var file = path.resolve(dir, 'app.js');
       var app = fs.readFileSync(file, 'utf8');
       assert.notEqual(app.indexOf('app.use(require(\'less-middleware\')(path.join(__dirname, \'public\')));'), -1);
@@ -305,7 +305,7 @@ describe('express(1)', function () {
       });
     });
   });
-  
+
   describe('--hbs', function () {
     var dir;
     var files;
--- a/test/cmd.js
+++ b/test/cmd.js
@@ -6,7 +6,6 @@ var mocha = require('mocha');
 var path = require('path');
 var request = require('supertest');
 var rimraf = require('rimraf');
-var async = require('async');
 var spawn = require('child_process').spawn;
 
 var binPath = path.resolve(__dirname, '../bin/express');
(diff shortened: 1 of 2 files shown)


test/cmd.js
```diff
@@ -335,35 +381,6 @@ describe('express(1)', function () {
       .expect(404, /<h1>Not Found<\/h1>/, done);
     });
   });
-
-  describe('--help', function () {
```
jordi12100: We had 2 tests with the same logic. -h and --help.

dougwilson: Sure, but now how do we know that -h still works? Looks like there is no longer a -h test.

jordi12100: Hmm, maybe we should keep both.. Now using async to handle both -h and --help without duplicating 30 lines of code.

dougwilson: I really want to accept the pull request, but I can't as long as we keep messing with the help test. Now it tests both, but when one is broken, you can't tell if -h is broken or --help is broken based on the test output and line numbers...

Please just keep all the help tests as-is for now. Perhaps make another pull request with proposed changes here, so I can accept the other parts of this pull request that matches the title.

jordi12100: No problem. It's now below --help as a separate test.

dougwilson: So the tests right now are (should) be in alphabetical order. I think that "h" comes alphabetically before "help", no?

jordi12100: Sorry, wasn't aware of that. -h comes before --help yes. 

-h was the last test in the file , so they weren't in alphabetic order.

dougwilson: It was in order as the last item, because "-h" comes last, since all the others have two dashes. "--" always comes before "-<letter>". If you put the descriptions into a JavaScript array and call .sort(), ideally it should match the order in the file.

dougwilson: The diff is really confusing right now, so I'm not even 100% sure. Basically when I look at the diff on this pull request, I expect to see "add less tests" and not "add less tests and move a bunch of things around". It makes it much easier to accept the pull request when the only changes are exactly the description of the pull request and not sneaking in a bunch of other things.

dougwilson: And to be clear: I'm not exactly rejecting the other changes as a whole, but it does mean that in order to even accept the less tests, we have to finish discussion on the entire pull request, so all the time spent discussion the help stuff is all the time I have to continue to not accept the new less tests. Keeping them as separate tests allows for parallel discussions and the acceptance of the things separately. I hope that makes sense.

jordi12100: Totally. Let's just stick to the purpose. Should be more clear now.

PR #98, opened 2016-01-14
upgrade hbs

yalishizhude:
hbs@3 has a bug when parsing **Inline Partials** and **Partial Blocks**,like:
```
{{#> myPartial }}
  Failover content
{{/myPartial}}
```
please update to version 4.0.0,thank you~


dougwilson:
Hi & thank you! We will definately add it in the next minor release of the generator :)

yalishizhude:
nice~~

Issue #90, opened 2015-12-08
Add support for dust template engine


nbro:
Dust is quite used template engine nowadays, and it would be nice to be able to have a generated with with our preferred engine. Is it possible to add it?

This generator is nice, but it lacks of many options. A generator would be really useful if customisable, IMO.

dougwilson:
Hi! Both are great suggestions! We'll put the adding of `dust` on our backlog to get added; if you have time and want to contribute a pull request adding it, that would be amazing, especially since we have not used that templating, so will also need to understand it enough to make some templates in it as well.

As for customization, that sounds good too! Can you provide some specifics on exactly what "customisable" means to you and what this module would look like if it were this way?

dossan:
I could for sure provide simple Dust templates as the ones provided for example for jade. By customisable I mean for example the possibility to specify the modules, maybe by having a config file so that we don't have to specify them every time, or something like that.You could also introduce an interface that guides the user in the creation of the app. These are just some ideas, but of course some might not be so useful at the end..

dougwilson:
Re: sample Dust templates, that would be very helpful!

Re customization, is that similar to the proposal in #80 ?

dossan:
For providing the templates, should I fork the repo and then add them to the `templates` folder?

Yes, something like that...

dougwilson:
> For providing the templates, should I fork the repo and then add them to the templates folder?

Sure, that works :) You can add the link to the commit here. You can even make it as a PR, even if incomplete and I can pick it from there. Whatever you think works best :)

> Yes, something like that...

Cool! If you have any additional thoughts to add to that issue, please feel free :)

loganhuskins:
Still looking for Dust templates? I'd be more than happy to do that for the project.

dougwilson:
@loganhuskins absolutely! You can find the most recent conversation on this at https://github.com/expressjs/generator/pull/121#issuecomment-255420551

dougwilson:
It's finally landing in the 4.15 version of the generator !

PR #104, opened 2016-03-14
Add support for dust template engine

hacksparrow:
This PR adds support for dust template engine https://github.com/expressjs/generator/issues/90.

The implementation is a little different from other template engines due to the fact that we have to use [`dustjs-linkedin`](https://www.npmjs.com/package/dustjs-linkedin), instead of `dust` (abandoned) or `dustjs`.

`dustjs-linkedin` does not provide a `__express()`-like method for rendering the templates, so we have to use a wrapper module, [`adaro`](https://www.npmjs.com/package/adaro).


.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules/
 temp/
 npm-debug.log
+.DS_Store
```
dougwilson: This is something that should be in your global `.gitignore`, not in each and every project's. Please remove.

README.md
```diff
@@ -30,7 +30,7 @@ Install dependencies:
 $ npm install
 ```
 
-Rock and Roll
+Start the app:
```
dougwilson: Please use a different PR for this change, rather than bundling with an unrelated change.

.gitignore
hacksparrow: Sure.

bin/express
```diff
@@ -185,6 +190,12 @@ function createApplication(app_name, path) {
     // Template support
     app = app.replace('{views}', program.template);
 
+    // Conditionally remove dust.js template code
+    if (program.template !== 'dust') {
+      app = app.replace("var adaro = require('adaro');\n", '');
```
dougwilson: This is going to be fragile. Can it be done similar to how the CSS stuff is done, where this is added, not removed?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "express-generator",
   "description": "Express' application generator",
-  "version": "4.13.1",
+  "version": "4.14.0",
```
dougwilson: Please don't bump the version in a pull request.

README.md
hacksparrow: Wasn't sure about this actually. Thanks for clarifying.

bin/express
hacksparrow: Will look into it.

package.json
hacksparrow: Ok. We should add this fact to the contribution guidelines doc.

/cc @tunniclm

dougwilson: It is common etiquette that a pull request never bumps a version, ever. Not sure why that would need to be spelled out. The pull request is only to add features, etc., not to dictate release schedules. Bumping the version here is missing a lot of things, for example, the version of this module follows the version on Express. I don't see this pull request bumping the Express dep to match the new version, and there are many other things that need to be done in this module before just bumping the version.

Are you not able to remove the version bump from the pull request?

dougwilson: There is not even a 4.14 line of Express yet, so bumping wouldn't even make sense (and version bumps are done in their own commit, when a version is cut @tunniclm already wrote about this in the pending release documentation).

hacksparrow: > Are you not able to remove the version bump from the pull request?

I updated it on my local repo a minute or so after your comment, it's just not pushed.

> There is not even a 4.14 line of Express yet ...

I was not aware about them being needing to be in sync, thanks for letting me know.

hacksparrow:
@dougwilson I have made the changes as per you suggestion.

dougwilson:

hacksparrow:
Something crazy going on in here! I can see some totally unrelated comments from 2014 and 2015. Might be best not to do anything till the dust settles.

dougwilson:
Re-opened as #121, thanks, @hacksparrow!

PR #105, opened 2016-03-22
Add newline to end of package.json

TrevorSayre:
- The file `package.json` is the only generated file that does not end in a newline
- POSIX [defines a line](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206) as terminating in a newline character
- `git diff` will warn when there is no newline at the end of a file
- GitHub shows a red icon at the end of a file when no newline is present
image

For these reasons, adding a newline at the end of `package.json` is a trivial fix to a common headache.

### DIFF
--- a/bin/express
+++ b/bin/express
@@ -238,7 +238,7 @@ function createApplication(app_name, path) {
     pkg.dependencies = sortedObject(pkg.dependencies);
 
     // write files
-    write(path + '/package.json', JSON.stringify(pkg, null, 2));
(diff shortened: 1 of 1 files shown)


dougwilson:
Sounds like a good change! All the tests are failing because this change breaks the tests. Can you update the tests to accommodate this change?

TrevorSayre:
Thank you, I'll look in to updating the tests soon.

PR #119, opened 2016-07-01
Add nunjucks engine support

### LINKED ISSUE #87: Add switch --nunjucks to add the Nunjucks template engine
This is by far my favorite template engine and it would be pretty sweet if this generator could automatically wire this up as the default template engine.

https://mozilla.github.io/nunjucks/

It is pretty simple to configure express.js to use nunjucks:

```javascript
var app = express();

nunjucks.configure('views', {
    autoescape: true,
    express: app
});

app.get('/', function(req, res) {
    res.render('index.html');
});
```
(above snippet taken from https://mozilla.github.io/nunjucks/getting-started.html)

gokaygurcan:
You can find more details about **nunjucks** from [here](https://mozilla.github.io/nunjucks/), [here](https://github.com/mozilla/nunjucks) and [here](https://www.npmjs.com/package/nunjucks).

It's just another template engine which has couple of good features. 

I needed to touch app.js to call nunjucks' _configure_ method, and you can see the reason [here](https://mozilla.github.io/nunjucks/getting-started.html). 

There was no test for EJS or Hogan.js, so I also didn't add any test. But if necessary, I can update this PR and add tests as well.

Closes #87 

P.S: Since no-one can open pr 103, I updated the nunjucks version and created this pr.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ This generator can also be further configured with the following command line fl
     -e, --ejs           add ejs engine support (defaults to jade)
         --hbs           add handlebars engine support
     -H, --hogan         add hogan.js engine support
+    -n, --nunjucks      add nunjucks engine support
     -c, --css <engine>  add stylesheet <engine> support (less|stylus|compass|sass) (defaults to plain css)
         --git           add .gitignore
     -f, --force         force on non-empty directory
--- a/bin/express
+++ b/bin/express
@@ -30,6 +30,7 @@ program
   .option('-e, --ejs', 'add ejs engine support (defaults to jade)')
   .option('    --hbs', 'add handlebars engine support')
   .option('-H, --hogan', 'add hogan.js engine support')
+  .option('-n, --nunjucks', 'add nunjucks engine support')
   .option('-c, --css <engine>', 'add stylesheet <engine> support (less|stylus|compass|sass) (defaults to plain css)')
   .option('    --git', 'add .gitignore')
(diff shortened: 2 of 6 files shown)


dougwilson:
I know this has sat around for a long time, and I'm very sorry! I dug into this module, mainly because the way it is attached smells weird, and I'm not sure if we should add nunjucks, at least using the module in this PR. The main issue is that it is overriding the `View` class on the Express.js app and does so in such a way that it's not possible to use multiple view engines in the same app, as it will simply ignore the set view engines and the file extensions (at least, this is my understanding from the source code and some testing).

gokaygurcan:
Aha! No worries, really. 
I was checking @adonisjs back then, and I saw that they are using it.

> AdonisJs has a solid View engine built on top of nunjucks. Everything from nunjucks is fully supported with some extra features unique to AdonisJs only.

And then I said "whatever, let's try this one in Express" and opened the PR. I really don't mind if you won't accept it because yea, it kinda overwrites some stuff in a bad way. 

Side-note: I never used nunjucks in a real product/application/website so far.

dougwilson:
Ah, I see. I was also just browsing their issues and it looks like they don't really want to deal with the Express.js integration anyway: https://github.com/mozilla/nunjucks/issues/611

dougwilson:
Thanks again for the PR, though! It is perfectly fine, the only real reason I'm even closing it is that it seems like the used module is really weird as-is and the maintainers don't even want to bother with the Express/js integration.

PR #117, opened 2016-06-30
Release 4.14

dougwilson:
This is a tracking issue for release 4.14.

**Please keep feature requests in their own issues**

If you want to make a comment on a particular change, please make the comment in the "Files changed" tab so comments are not lost during a rebase.

List of changes for release:
- [x] Add `pug` as a view engine #107 #109 #110 #113 #114 #115 #116 #118
- [x] Add `twig` as a view engine #100 #122
- [x] Add `vash` as a view engine #124
- [x] Add warning when using the default view engine #141
- [x] Create valid package name #143
- [x] Move view choices under a `--view` option #44 #92 #120 #128
- [x] Provide help/usage when `--css` missing `<engine>`
- [x] Remove `-V` as an alias to `--version`
- [x] Rename variable holding main router #131
- [x] Show help/usage when given an unknown option
- [x] Simplify the default error handler #70

**Testing this release**

If you want to try out this release, you can install it with the following commands:
```bash
$ npm cache clean express-generator
$ npm install expressjs/generator#4.14
```

Owners/collaborators: please do not merge this PR :)

PR #120, opened 2016-07-01
Make the template option -v accept 4 templates

azat-co:
The new option `-v` will accept 4 template engines and error on the unknown input. #117 #92  @dougwilson


dougwilson:
This has _finally_ made it into the pending 4.14 changes, without the removal of the existing switches to make it easier to transition. Thanks again for all your hard work!

PR #121, opened 2016-07-01
Adds support for dust template engine

hacksparrow:
This PR adds support for dust template engine https://github.com/expressjs/generator/issues/90.

The implementation is a little different from other template engines due to the fact that we have to use [`dustjs-linkedin`](https://www.npmjs.com/package/dustjs-linkedin), instead of `dust` (abandoned) or `dustjs`.

`dustjs-linkedin` does not provide a `__express()`-like method for rendering the templates, so we have to use a wrapper module, [`adaro`](https://www.npmjs.com/package/adaro).

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ This generator can also be further configured with the following command line fl
     -e, --ejs           add ejs engine support (defaults to jade)
         --hbs           add handlebars engine support
     -H, --hogan         add hogan.js engine support
+    -d, --dust          add dust.js engine support
     -c, --css <engine>  add stylesheet <engine> support (less|stylus|compass|sass) (defaults to plain css)
         --git           add .gitignore
     -f, --force         force on non-empty directory
--- a/bin/express
+++ b/bin/express
@@ -30,6 +30,7 @@ program
   .option('-e, --ejs', 'add ejs engine support (defaults to jade)')
   .option('    --hbs', 'add handlebars engine support')
   .option('-H, --hogan', 'add hogan.js engine support')
+  .option('-d, --dust', 'add dust.js engine support')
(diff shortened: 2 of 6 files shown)


dougwilson:
So since this wasn't merged right away (sorry!) I did go and check out what the current dust + express landscape looked like on npm to see if there is a simple way to just use one module. I found https://github.com/raxell/dustjs-express which looks like it may be good, and even recently created. Looks to support the array of views introduced in Express recently as well. Thoughts on that module? (I don't know much about dust).

loganhuskins:
It looks like [adaro](https://www.npmjs.com/package/adaro) is a little more active than [dustjs-express](https://github.com/raxell/dustjs-express) is. I can change this over to use [dustjs-express](https://github.com/raxell/dustjs-express) if that's what you'd prefer, but if [adaro](https://www.npmjs.com/package/adaro) is fine I think it may be better to stick with that.

If that's fine, I can go through and resolve any conflicts so it can be merged in.

dougwilson:
At the very least, the module should be providing the special ~__express~ function or one that works as close as possible, otherwise all the code to get that to work will just end up in the generator, completely unreusable to the rest of the community, and would be a maintain an burden here. It should integrate just as easily as all the other template engines do without workarounds, otherwise it isn't in a state to be included in the generator.

dougwilson:
So this PR has been around for quite some time and is difficult to merge, especially since the way new view engines are added has changed in the interim. If someone wants to re-open it with the latest semantics, that'd be great. I have a WIP branch that templates the files with `ejs` (https://github.com/expressjs/generator/tree/templated) that could make using `adaro` and similar wrappers even easier to add.

dougwilson:
So it turns out after I just finished and merged in the templated branch, it was not too difficult to fix up this PR for 4.15, so @hacksparrow I'm landing this PR on 4.15 :)

Issue #133, opened 2016-09-04
Default template engine is still Jade, rather than Pug


stuartellis:
By default, the current release generates projects with Jade, rather than the replacement, Pug. This is a legal (copyright) issue, more than a technical one:

https://github.com/pugjs/pug/issues/2184

dcollao:
![image](https://cloud.githubusercontent.com/assets/20039386/18341898/9d9f8ba2-7582-11e6-8ccf-f86749b82b86.png)

blakeembrey:
I've seen this issue logged every other week. Please search for existing issues before opening another one. E.g. https://github.com/expressjs/generator/issues?q=is%3Aissue+jade+is%3Aclosed, there's a couple of already closed issues with descriptions.

PR #134, opened 2016-09-07
PR - Default template engine is still Jade, rather than Pug #133

osobh:
I forked this repo then addressed the remaining references to the past Jade name and changed them to Pug. I then ran the tests and ran into some issues with Mocha and using mocha.before and mocha.after. Ready for Feedback on these changes.


dougwilson:
1. Please change your PR to be against the 4.14 branch instead of master, or I cannot merge it due to conflicts.
2. I think there are similar PRs?
3. I don't think it makes sense to default people to an unfinished beta project. If we are going to change from jade to something else, let's use a project that has some non beta version.

osobh:
sounds good, we can close this and I can resend a new one when pug is out of beta. Thank you Doug!

PR #144, opened 2016-11-03
Dynamically record application routes

hiagodotme:
Dynamically record application routes

### DIFF
--- a/templates/js/app.js
+++ b/templates/js/app.js
@@ -6,7 +6,7 @@ var cookieParser = require('cookie-parser');
 var bodyParser = require('body-parser');
 var fs = require('fs');
 var routesDir = './routes/';
-
+var registerLog = false;
 var app = express();
 
 // view engine setup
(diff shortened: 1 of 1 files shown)


hiagodotme:
Hello I do not think it's cool I have to keep logging the routes, so I'm assuming that everything in the routes folder is a route. So I made an automatic importer.

hiagodotme:
Honestly, I do not know why you're not testing.

gokaygurcan:
I think everyone has their own way of registering routes. One may use one big routes.js while another uses bunch of small or logically separated files. 
At this point, generator doesn't try to direct you. Like Express' itself, I like to have a generator which is un-opinionated. 

This changes forces me to have a `./routes` folder, have an `index.js` file under that folder and have all the other files under routes folder again. I don't like this kind of approaches. 

```javascript
var index = require('./routes/index');
app.use('/', index);
```

This is simple, I can just change the path in require and voilà! In this new usage with dynamic routes (!), I need to change the whole logic. And finally, there're these performance problems too, having two synchronous method calls and a loop. 

Of course, this is just my idea, I'd like to hear what others think.

dougwilson:
So the reason the tests are failing is because the changes will no longer properly work when the process cwd is not the directory with the app, which is a common situation in CIs (like Travis CI) and app SaaS environments. I haven't heard from any other Express members on this, but my opinion on this is twofold: (1) this is just too much logic being added to the generator's code directly that needs to be maintained by the express-generator project (instead of pulling in a dependency) and (2) I think it is too opinionated, especially with the `app.use` path being determined by the name of a file on the file system, which means that you're bring in the character limits and case sensitivity limits of the underlying file system into the routing and limits you only to strings for the paths.

PR #146, opened 2016-12-23
Add ES2015 syntax to v4.14

asimpletune:
I noticed that the other PR for this was closed and a label "needs rebase" was added, so I merged my previous changes with he most recent ones from this project. Didn't take long, as I had already merged these changes back when there was a "4.14" branch. Thanks!


test/cmd.js
```diff
@@ -14,12 +14,12 @@ var TEMP_DIR = path.resolve(__dirname, '..', 'temp', String(process.pid + Math.r
...
 describe('express(1)', function () {
   before(function (done) {
-    this.timeout(30000);
+    this.timeout(60000);
     cleanup(done);
   });
 
   after(function (done) {
-    this.timeout(30000);
+    this.timeout(120000);
```
dougwilson: These timeouts are large as they are and now even larger. Please revert them back.

test/cmd.js
```diff
@@ -115,7 +115,7 @@ describe('express(1)', function () {
       it('should create basic app', function (done) {
         run(ctx.dir, [], function (err, output) {
           if (err) return done(err)
-          assert.equal(parseCreatedFiles(output, ctx.dir).length, 17)
```
dougwilson: What is the extra file that is being added to all the non-es2015 apps?

test/cmd.js
asimpletune: I remember that I raised the timeouts previously, because the AppVeyor tests would unpredictably timeout - if I recall the `--es2015` tests would vary greatly in duration - so bearing that in mind let me know if there's anything different that comes to mind that you'd like me to do. 

It looks like they passed this submission so maybe this issue has been resolved already.

test/cmd.js
asimpletune: The extra file is added to more elegantly support the transpiling. I recall this issue was raised earlier, although I can't find the exact comment, but the suggested solution was to break out the extra file. 

I think it may make more sense to explain by just showing the result of running `express --es2015` 
example

I vaguely remember the original concern was that the resulting generated `www` file didn't use the es2015 syntax itself, due to how `babel-core/register` [works](https://babeljs.io/docs/usage/babel-register/). So, the idea was to keep `bin/www` as the entrypoint to the application, but breaking the actual server code into into `server.js`, which is located in the root dir of the generated app.

dougwilson: Why would that affect the non-es2015 apps?

asimpletune: I wrote it that way for the sake of consistency. I don't think it removes the clarity of what's going - it's clear both ways - whether `bin/www` serves sort of as a `main()`(we know or what those does by convention), or you go to make changes to the server (as an abstract concept) at `server.js`. However, I think changing the file structure just when using es2015 syntax, and keeping it as is otherwise, could actually reduce clarity in a lot of cases. 

Additionally, I think it's not too idiosyncratic to have a single file, in this case `www`, that simply serves  as an entrypoint to all the things that need to be done at startup, especially if it's located underneath `bin/` (although I could be wrong, I don't have as much experience as you so please let me know. I've seen a lot of things like this, where command parsing and side effects are done in a single file that's the actual entry point but doesn't do anything else). Previously, only one thing had to be "done", i.e. start the server, but now two thing needs to be done, i.e. transpile the code, then start the server. In the future, perhaps more things will need to be done upon startup that don't involve starting the server or transpiling the code? FWIW, personally, if I'm trying to understand some new code, I usually look for such a file to follow the "signal flow", so to speak. 

So, intuitively, it felt like, given the circumstances of how `babel-core/register` works, no way around that aspect of the problem, that I would be adding complexity by doing the alternative, where as the way I have it here was consistent and still basically just as concise. Although obviously this up to your discretion, I'm happy to do whatever works best for you and the project.

dougwilson: I think there is to much discussion that needs to happen i.r.t to the `server.js` and `www` split. Can you please pull that out of this PR and into it's own PR for a discussion?

asimpletune: You bet

lewis-yuburi: Great work and iniciative @asimpletune!! I want to contribute!
Btw, for `es6/es2015` - `babel` we need to transpile the code to `es5` (nodejs compatible) because `babel-node` and `babel-core/register` aren't for production usage (see [https://github.com/babel/example-node-server/issues/13](see)). More info [here](https://github.com/babel/example-node-server#getting-ready-for-production-use), and [here](https://medium.com/@Cuadraman/how-to-use-babel-for-production-5b95e7323c2f#.bwjjho9m2).

asimpletune: Thanks @lewis-yuburi, I will look into including the changes you referenced in the example.

dougwilson:
Hi @asimpletune the PR no longer easily merges and it doesn't seem like there have been any changes from our conversations above, so it seems stalled out. Please feel free to re-open the PR when you get a chance to address things, especially something around the server vs www split (as a separate PR or issue).

asimpletune:
Ok, thanks for the heads up @dougwilson, and sorry about that.

dougwilson:
Also, the templating has been overhauled to use EJS instead of trying to do string replaces in the file, which may make this an easier task now.

asimpletune:
Good to know, I can't wait to see the new code!

--  
-Spence

dougwilson:
@asimpletune merged into `master` now :)

nyteshade:
This isn't reflected in the README.md. Can you add?

dougwilson:
Hi @nyteshade , reading the last few comments, I see how it's confusing, sorry! So what happened here is the changes for ES2015 sat around for a long time, was stale and I couldn't merge. Since there wasn't any activity after my comments, I just closed this PR for another one to get remade if someone was so inclined.

The merging into `master` above was actually regarding the internal changes to use EJS templates to create the generated files, which should make it even easier for whoever wants to pick up adding ES2015 syntax as an option to the generator.

I hope that makes sense!

PR #148, opened 2016-12-24
Use const

pravdomil:
ES6

### DIFF
--- a/templates/js/www
+++ b/templates/js/www
@@ -4,22 +4,22 @@
  * Module dependencies.
  */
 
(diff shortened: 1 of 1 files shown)


dougwilson:
"var" is just as valid as "const". The only thing that would yell at you is if you are using a linter, which you can just ignite this file.

dougwilson:
I also believe this is not backwards compatible, since I don't think const exists in Node.js 0.10 which this project generates for as a supported runtime.

Issue #152, opened 2017-02-18
Option for no view engine


tobymurray:
With the wealth of front end frameworks available now, I don't find it appealing to use any of the view engines, though bootstrapping technologies like this project appeal to me. My use case is as follows:

- Angular front end transpiled/compiled/obfuscated/minified/whatever into a bunch of static blobs
- Nginx delivering the client (bunch of static blobs)
- Express providing the API for the client and all the other goodness.

What are your thoughts on providing another flag (e.g. --view none) to avoid the view engine entirely? Perhaps if the same "plug-and-play" experience is desired the routes could serve up regular old HTML files from the public folder? 

I can take a stab at adding it if it's viable.

dougwilson:
I think that is a good idea.

tobymurray:
I think I've got something close to working, but it's just a tiny bit awkward because "none" is different than every other view engine - e.g. the `// view engine setup` section doesn't need to be there. Looking at #141 - what are your thoughts about defaulting to no view engine? I think conceptually it makes sense, but it's definitely a change from the status quo.

dougwilson:
Yea, defaulting is certainly a discussion, but we can add it even without it being the default. A few PRs tried this previously, but got abandoned. The called it "static" instead of none, which is probably a good name for the engine.

PR #153, opened 2017-02-24
#152: Add option for no view engine

tobymurray:
Add '--none' flag to avoid installing any view engine. Serve static HTML files in place of the index and error views.

### DIFF
--- a/bin/express
+++ b/bin/express
@@ -81,14 +81,14 @@ function addIndexServing(app, viewEngine) {
 function addViewEngine(app, viewEngine) {
   if (program.view == 'none') {
     app.locals.viewEngineSetUp = '';
-    app.locals.renderError = "res.sendFile('error.html', {\n" +
-      "    root: __dirname + '/public/'\n" +
+    app.locals.renderError = "res.sendFile('error.html', {" + eol + 
+      "    root: __dirname + '/public/'" + eol + 
       "  });";
   } else {
-    app.locals.viewEngineSetUp = "\n" +
-      "// view engine setup\n" +
-      "app.set('views', path.join(__dirname, 'views'));\n" +
-      "app.set('view engine', '" + viewEngine + "');\n";
+    app.locals.viewEngineSetUp = eol +
+      "// view engine setup" + eol +
+      "app.set('views', path.join(__dirname, 'views'));" + eol +
+      "app.set('view engine', '" + viewEngine + "');" + eol;
     app.locals.renderError = "res.render('error');";
   }
 }
@@ -264,7 +264,13 @@ function createApplication(app_name, path) {
         app.locals.css = eol + 'app.use(require(\'node-compass\')({mode: \'expanded\'}));'
         break;
       case 'sass':
-        app.locals.css = eol + 'app.use(require(\'node-sass-middleware\')({\n  src: path.join(__dirname, \'public\'),\n  dest: path.join(__dirname, \'public\'),\n  indentedSyntax: true,\n  sourceMap: true\n}));'
+        app.locals.css = eol + 
+          'app.use(require(\'node-sass-middleware\')({' + eol + 
+          '  src: path.join(__dirname, \'public\'),' + eol + 
+          '  dest: path.join(__dirname, \'public\'),' + eol + 
+          '  indentedSyntax: true,' + eol + 
+          '  sourceMap: true' + eol + 
+          '}));'
         break;
     }
 
@@ -339,7 +345,7 @@ function createApplication(app_name, path) {
     pkg.dependencies = sortedObject(pkg.dependencies);
 
     // write files
-    write(path + '/package.json', JSON.stringify(pkg, null, 2) + '\n');
+    write(path + '/package.json', JSON.stringify(pkg, null, 2) + eol);
     write(path + '/app.js', app.render())
     mkdir(path + '/bin', function(){
       write(path + '/bin/www', www.render(), 0755)
@@ -472,7 +478,7 @@ function main() {
   
   // Default view engine (a value of true here indicates the view engine has not been otherwise specified)
   if (program.view === true) {
-    warning("the default view engine will not be jade in future releases\nuse `--view=jade' or `--help' for additional options")
+    warning("the default view engine will not be jade in future releases" + eol + "use `--view=jade' or `--help' for additional options")
     program.view = 'jade'
   }
 
@@ -516,7 +522,7 @@ function renamedOption(originalName, newName) {
 
 function warning(message) {
   console.error()
-  message.split('\n').forEach(function (line) {
+  message.split(eol).forEach(function (line) {
     console.error('  warning: %s', line)
   })
   console.error()
--- a/bin/express
+++ b/bin/express
@@ -50,6 +50,7 @@ program
   .option('    --hbs', 'add handlebars engine support', renamedOption('--hbs', '--view=hbs'))
   .option('-H, --hogan', 'add hogan.js engine support', renamedOption('--hogan', '--view=hogan'))
   .option('-v, --view <engine>', 'add view <engine> support (ejs|hbs|hjs|jade|pug|twig|vash) (defaults to jade)')
+  .option('    --no-view', 'use HTML directly instead of a view engine')
   .option('-c, --css <engine>', 'add stylesheet <engine> support (less|stylus|compass|sass) (defaults to plain css)')
   .option('    --git', 'add .gitignore')
   .option('-f, --force', 'force on non-empty directory')
@@ -57,6 +58,39 @@ program
 
 if (!exit.exited) {
   main();
+}
+
+/**
+ * Replace placeholders for serving index page
+ */
+
+function addIndexServing(app, viewEngine) {
+  if (program.view == 'none') {
+    app.locals.requireIndexRoute = '';
+    app.locals.indexRoute = '';
+  } else {
+    app.locals.requireIndexRoute = eol + "var index = require('./routes/index');";
+    app.locals.indexRoute = eol + "app.use('/', index);";
+  }
+}
+
+/**
+ * Replace the template placeholders according to the configured view engine
+ */
+
+function addViewEngine(app, viewEngine) {
+  if (program.view == 'none') {
+    app.locals.viewEngineSetUp = '';
+    app.locals.renderError = "res.sendFile('error.html', {\n" +
+      "    root: __dirname + '/public/'\n" +
+      "  });";
+  } else {
+    app.locals.viewEngineSetUp = "\n" +
+      "// view engine setup\n" +
+      "app.set('views', path.join(__dirname, 'views'));\n" +
+      "app.set('view engine', '" + viewEngine + "');\n";
+    app.locals.renderError = "res.render('error');";
+  }
 }
 
 /**
@@ -134,6 +168,8 @@ function createApplication(app_name, path) {
   // JavaScript
   var app = loadTemplate('js/app.js');
   var www = loadTemplate('js/www');
+  var index = program.view == 'none' ? null : loadTemplate('js/routes/index.js');
+  var users = loadTemplate('js/routes/users.js');
 
   // App name
   www.locals.name = app_name
@@ -165,8 +201,11 @@ function createApplication(app_name, path) {
     });
 
     mkdir(path + '/routes', function(){
-      copy_template('js/routes/index.js', path + '/routes/index.js')
-      copy_template('js/routes/users.js', path + '/routes/users.js')
+      if (program.view != 'none') {
+        copy_template('js/routes/index.js', path + '/routes/index.js') 
+      }
+
+      copy_template('js/routes/users.js', path + '/routes/users.js') 
       complete();
     });
 
@@ -205,6 +244,10 @@ function createApplication(app_name, path) {
           copy_template('vash/layout.vash', path + '/views/layout.vash');
           copy_template('vash/error.vash', path + '/views/error.vash');
           break;
+        case 'none':
+          copy_template('none/index.html', path + '/public/index.html');
+          copy_template('none/error.html', path + '/public/error.html');
+          break;
       }
       complete();
     });
@@ -226,7 +269,10 @@ function createApplication(app_name, path) {
     }
 
     // Template support
-    app.locals.view = program.view
+    addViewEngine(app, program.view);
+
+    // Index route
+    addIndexServing(app, program.view);
 
     // package.json
     var pkg = {
@@ -265,6 +311,9 @@ function createApplication(app_name, path) {
         break;
       case 'vash':
         pkg.dependencies['vash'] = '~0.12.2';
+        break;
+      case 'none':
+        // No additional dependencies needed
         break;
       default:
     }
@@ -309,6 +358,7 @@ function copy_template(from, to) {
   from = path.join(__dirname, '..', 'templates', from);
   write(to, fs.readFileSync(from, 'utf-8'));
 }
+
 /**
  * Create an app name from a directory path, fitting npm naming requirements.
  *
@@ -372,6 +422,14 @@ function launchedFromCmd() {
 }
 
 /**
+ * Determine whether legacy view syntax (e.g. '--ejs') was used to specify view
+ */
+
+function legacyViewSpecified() {
+  return program.ejs || program.hbs || program.hogan || program.pug;
+}
+
+/**
  * Load template file.
  */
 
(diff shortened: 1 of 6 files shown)


tobymurray:
Hmm... 0.10 and 0.12 support? They're both past their EOL now that we're in 2017 (see [Node doc](https://github.com/nodejs/LTS)), is this project intending to support those versions for the foreseeable future?

dougwilson:
Yea, we support all the versions that Express supports.

bin/express
```diff
@@ -49,7 +49,7 @@ program
   .option('    --pug', 'add pug engine support', renamedOption('--pug', '--view=pug'))
   .option('    --hbs', 'add handlebars engine support', renamedOption('--hbs', '--view=hbs'))
   .option('-H, --hogan', 'add hogan.js engine support', renamedOption('--hogan', '--view=hogan'))
-  .option('-v, --view <engine>', 'add view <engine> support (ejs|hbs|hjs|jade|pug|twig|vash) (defaults to jade)')
+  .option('-v, --view <engine>', 'add view <engine> support (ejs|hbs|hjs|jade|pug|twig|vash|none) (defaults to jade)')
```
dougwilson: Previous PRs called it "static" which seems like a nice name. At the very least, this list is alphabetical.

bin/express
```diff
@@ -312,10 +319,27 @@ function createApplication(app_name, path) {
...
 
+function addViewEngine(app, viewEngine) {
+  if (program.view == 'none') {
+    app = app.replace('{view-engine-set-up}', '');
+    return app.replace('{render-error}', `res.sendFile('error.html', {
+    root: __dirname + '/public/'
+  });`);
+  } else {
+    app = app.replace('{view-engine-set-up}', `
+// view engine setup
```
dougwilson: Can the indenting get fixed here? Looks really weird with this offset indenting.

test/cmd.js
```diff
@@ -895,6 +895,60 @@ describe('express(1)', function () {
         .expect(404, /<h1>Not Found<\/h1>/, done)
       })
     })
+
+    describe('--none', function () {
```
dougwilson: Looks like the test is misnamed; the args in here are `--view=none`, not `--none`.

templates/none/routes/index.js
```diff
@@ -0,0 +1,10 @@
+var express = require('express');
+var router = express.Router();
+
+router.get('/', function (req, res, next) {
+  res.sendFile('index.html', {
```
dougwilson: This is just a general comment for discussion, not a request to change : does using `res.sendFile` make sense here, so should it just be let up to `express.static`?

bin/express
tobymurray: Are you saying you prefer `static` over `none`? Does it make sense to describe this as having a static view engine? My understanding is the "engine" part is anything above and beyond HTML - are there any "engine" features available with only static content?

Can definitely make this alphabetical - didn't realize that was the case.

templates/none/routes/index.js
tobymurray: I don't have much of an opinion - I'm just starting in with Express, so I was trying to mimic the existing view engines. If using `express.static` this could just be folded into `app.js`, right?

tobymurray:
> Yea, we support all the versions that Express supports.

That blows a little bit... I'll take a look at making this work in 0.10.

bin/express
dougwilson: I honestly don't have a preference. I can see arguments for when it was argued to be called `static` (a good reference to `express.static`), I can see arguments for `none`, and even calling it `html`.

templates/none/routes/index.js
dougwilson: Yea, there just wouldn't be a `/` route at all, since `express.static` is enabled to check for `index.html` for a directory request by default.

dougwilson:
We plan on changing it in the next major version. If you're OK with your PR waiting to land until then, then you don't have to worry about 0.10.

bin/express
tobymurray: If you're happy either way, I like 'none' best as I think it's the most clear when talking about what view engine is there (I'd be similarly happy with HTML). The correlation to `express.static()` is interesting, but I think it's marginally less understandable - I don't think I'd have a good idea what a "static view engine" was.

templates/none/routes/index.js
tobymurray: I've left it for now. This change would make the `none` view engine more different than the rest of the view engines. I'm up for changing it if `express.static` is more desirable or more idiomatic Express.

dougwilson: It probably is, especially because I just tested this generated app and this function is ever called for `/`, since the existing `express.static` is serving it back. I think having this function when it doesn't even do anything is probably going to be super confusing...

test/cmd.js
```diff
@@ -895,6 +895,60 @@ describe('express(1)', function () {
         .expect(404, /<h1>Not Found<\/h1>/, done)
       })
     })
+
+    describe('--view=none', function () {
```
dougwilson: Probably should be moved up into the existing `--view <engine>` section and renamed to `none`

bin/express
```diff
@@ -312,10 +319,27 @@ function createApplication(app_name, path) {
   });
 }
 
+function addViewEngine(app, viewEngine) {
```
dougwilson: The new function should really get a jsdoc block like all the others. And in general, all the functions are organized in alpha-order from top to bottom. Not sure what the thought was for the placement of this function was.

bin/express
dougwilson: Yea, idk, it's more about what is less confusing to the general Express users. The only thing I really have is that all the previous issues / PR called it "static" and no one until this one thought to call it "none", so that's all I have to go on.

templates/none/routes/index.js
tobymurray: Ah, good call. I didn't realize that it wasn't getting evaluated - I'll remove it.

test/cmd.js
tobymurray: I believe this is already in the `--view <engine>` section, and renaming it would just be reverting would be almost the opposite of 83858da2

I can do that, but so far as I can tell it's just a naming thing originally `--none` -> `--view=none` -> `none`

dougwilson: Oh, oops, the build output was weird. Yea, it should be in alpha order like the test of the tests in the section and just named `none`, i.e. the section is the name of what is filled into `<engine>` placeholder.

tobymurray:
I'm still not keen on `static` (although I might just be misunderstanding things), but I'm happy with `html` if you'd prefer - no real opinion between the two.

dougwilson:
Oh! It just came to me: I wonder if we can make this --no-view which is the typical style to disable something.

tobymurray:
Whatever you'd like. `--no-view` as in a different flag from `--view`? Like that better than the others?

dougwilson:
Yea, something. I'm thinking `--no-view` as the disablement of `--view`, which is what you were arguing for with the conversation of `none` vs `static`. Just take a look at something like `git gc --help` for example of what is common in command line programs. For example, it has `--prune=<date>` or `--no-prune`. That seems similar to us having a `--view=<engine>` or `--no-view`, which makes sense to me, at least.

tobymurray:
Sounds reasonable to me, and keeps the options a little more sensible - separates the view engine options from the absence-of-view-engine option. I'll give it a shot this weekend.

tobymurray:
This looks like it works for me now with the `--no-view` options. It looks like commander supports that functionality directly.

bin/express
```diff
@@ -398,18 +448,20 @@ function main() {
   var appName = createAppName(path.resolve(destinationPath)) || 'hello-world'
 
   // View engine
-  if (program.view === undefined) {
+  if (program.view) {
```
dougwilson: I'm not sure if it's this line, but something around here broke the preference for `--view=<engine>` over the old `--<engine>` switches. Currently, if you call `express --ejs --view=pug`, we use the newer variant (pug is used). This PR reveres this for some reason and now causes the same command to generate with ejs.

tobymurray: Is the desired behavior to allow conflicting views to be specified and give preference to the more recent format? Would it be better to blow up, as it seems like it's obviously an error?

dougwilson: I could see both ways, I guess. So this PR would need to preserve the existing behavior. Blowing up may be good, so feel free to make a separate PR for that change, so the rest of the members can weight in on that change :) !

tobymurray:
I think I've resolved the regression, and I added a test to assert that.

test/cmd.js
```diff
@@ -343,6 +343,45 @@ describe('express(1)', function () {
     });
   });
 
+  describe('--ejs --view=pug', function () {
+    var ctx = setupTestEnvironment(this.fullTitle())
+
+    it('should create basic app with pug templates', function (done) {
+      run(ctx.dir, ['--view', 'pug'], function (err, stdout) {
```
dougwilson: Though the description of the tests says "--ejs --view=pug", the actual command being executed is not that, but simply "--view pug"

bin/express
```diff
@@ -398,18 +456,22 @@ function main() {
   var appName = createAppName(path.resolve(destinationPath)) || 'hello-world'
 
   // View engine
-  if (program.view === undefined) {
+  if (program.view && program.view !== true && legacyViewSpecified()) {
+    // Prefer new view syntax over legacy syntax, so ignore the legacy parameter
```
dougwilson: Can you invert the conditional such that it does not create an empty block?

bin/express
```diff
@@ -398,18 +456,22 @@ function main() {
...
+  } else if (!program.view) {
+    program.view = 'none'
+  } 
+  
+  // Default view engine (a value of true here indicates the view engine has not been otherwise specified)
+  if (program.view === true) {
     warning("the default view engine will not be jade in future releases\nuse `--view=jade' or `--help' for additional options")
     program.view = 'jade'
-  }
+  }
```
dougwilson: Accidental trailing white space added.

test/cmd.js
```diff
@@ -467,6 +506,60 @@ describe('express(1)', function () {
     })
   })
 
+  describe('--no-view', function () {
+    var ctx = setupTestEnvironment(this.fullTitle())
+
+    it('should create basic app with no view engine', function (done) {
+      run(ctx.dir, ['--view', 'none'], function (err, stdout) {
```
dougwilson: It looks like the command that is being executed does not match the description. I thought you removed --view=none in favor of --no-view. Why does the old command still work?

tobymurray: It works mostly by coincidence. `--no-view` ultimately maps to `program.view === 'none'`, but there's nothing that validates that `program.view` is set to a supported value with the `--view=<value>` parameter. So passing in `--view=none` also sets the `view` to `none`. I think the only way to avoid something like that occurring (if it were changed to `foo`, it would work for `--view=foo`) is with active validation that the parameter passed in is one of the suspected values. Perhaps it'd be marginally less surprising if I change `none` to `no-view`, although that pretty much just hides the issue.

tobymurray:
How have I managed to accumulate conflicts? So far as I can tell I'm still working off the head of master and I don't see how I could have introduced a conflict with myself...

bin/express
tobymurray: Whoops, cleared up in 3215dc7

bin/express
tobymurray: Inverted in 8871283

test/cmd.js
tobymurray: Yikes, that's sloppy. Good catch - corrected in b7d57c0

test/cmd.js
dougwilson: Ah, yes, I forgot that there was discussion a long time ago to make it validate, but that never got added.

dougwilson:
I think that may have been my bad. I just got on my main computer and went to this project in order to start merging your PR since we need to get a release out for Express 4.15.0 asap, and syncing the repo I think threw up some commits from last month. You're welcome to rebase (not merge) your changes, otherwise just leave as-is and I can work at rebasing them.

tobymurray:
Ah, I hadn't added this as upstream, that's why I wasn't seeing them. I'll give rebasing a shot.

tobymurray:
I believe this is ready for a re-review when you have a chance.

bin/express
```diff
@@ -60,6 +61,39 @@ if (!exit.exited) {
...
+}
+
+/**
+ * Replace the template placeholders according to the configured view engine
+ */
+
+function addViewEngine(app, viewEngine) {
+  if (program.view == 'none') {
+    app.locals.viewEngineSetUp = '';
+    app.locals.renderError = "res.sendFile('error.html', {\n" +
```
dougwilson: Above you're using the `eol` variable for the new line, but here just `\n`. Use one or the other?

dougwilson:
Was the `--view=none` thing ever resolved?

tobymurray:
> Was the --view=none thing ever resolved?

We talked about it [here](https://github.com/expressjs/generator/pull/153#discussion_r104339019). I assumed the validation was a separate issue. Perhaps something to address in a follow up?

dougwilson:
That's perfectly fine. Pleas let ping me when you've added the follow up commit here to address :)

bin/express
tobymurray: It seems like `eol` and `\n` are not used consistently elsewhere - is there a preference? I assume `eol`?

dougwilson: Yea, I noticed that somewhere else in there. I assume that it should be `eol` for Windows users, but I'm not sure if they are ended up with a LF file from the registry or a CRLF file...

tobymurray:
I replaced all the `'\n'` with `eol` in 642ffdb

tobymurray:
Newlines and strict string matching assertions look like a bit of a mess. In a09fae2 I tried to use the platform specific newlines, but it looks like it's not consistent (see failing tests in Windows). 1d9b82f is _an_ answer to that - using regexs instead. Hopefully that's workable, @dougwilson?

dougwilson:
Hi @tobymurray I just took a look at your updates and I'm a bit confused as to what is going on; my comments on the `\n` vs `eol` was only regarding the changes you were making, not the entire project. I think updating the entire project is good, that that would probably be more of something to do in a different PR, rather than postponing this one further. The only thing I've been waiting on is the fix to not accept `--view=none`. I'm not sure I'm seeing this in the changeset, have you addressed it?

tobymurray:
I think we've got our wires crossed here.

> my comments on the \n vs eol was only regarding the changes you were making

My bad on that one, I didn't see the value in consistency when the project was inconsistent, so I assumed you meant aligning the project. 

> The only thing I've been waiting on is the fix to not accept --view=none

That is what I was speaking to in my comment [here](https://github.com/expressjs/generator/pull/153#issuecomment-285906434). The validation to not accept arguments that happen to match the string representation is an existing issue - as you mentioned [here](https://github.com/expressjs/generator/pull/153#discussion_r104339019). 

I thought we were agreeing to punt the validation and you were waiting for newline consistency, so no, I have not addressed it. Clearly that's not the case.

dougwilson:
> That is what I was speaking to in my comment here. The validation to not accept arguments that happen to match the string representation is an existing issue - as you mentioned here.

So it sounds like you may have misunderstood or because you said a few different things in the paragraph, I wasn't clear enough in what I meant.

The "previous issue" is `--view=foo` does not error. I'm not expecting that to be fixed here.

But that is not the same as the `--view=none` issue. This PR needs to treat it exactly the same as `view=foo`, not the same as `--no-view`. That is the bug this PR is introducing and I'm expecting to be fixed.

So perhaps to help along, let's set out my expectations for the desired changes to this PR:

1. Back out all the `\n` => `eol` changes into another PR, otherwise you're just going to have to make even more changes here, because I see some issues regarding the changes you made there, but would rather focus on landing the `--no-view` feature, not endlessly expanding the scope of this PR.
2. Fix `--view=none` to behave exactly how `--view=foo` behaves.

Because this has taken so long, I'm going to release 4.15 of the generator tonight so people can start generating Express 4.15 apps. I was really hoping to land your feature in 4.15, since it's a new feature, but alas it has been almost a month now. Because the version number of this module are locked to the version number of Express, this means that missing the 4.15 release means I have no idea when this feature will land in the generator, since that would depend on when a 4.16 Express gets released.

tobymurray:
> But that is not the same as the --view=none issue

Makes sense.

> I have no idea when this feature will land in the generator

Also makes sense. For the sake of my own sanity, I'm going to close this - a month is just too long. If I get some time to tidy it up, I'll re-issue in a (hopefully) more production-ready state.

PR #158, opened 2017-04-28
#152: Add option for no view engine

tobymurray:
A rework of [this declined PR](https://github.com/expressjs/generator/pull/153).

Summary:
- Passing the `--no-view` option excludes the view engine set up and corresponding render calls, instead statically serving HTML
- I made an effort to avoid the pitfalls of the previous attempt by using `null` instead of `'none'`. This should remove the concerns around the previously erroneous `view=none` behavior, now treating that the same as `view=foo`.

With the recent addition of the lint configuration, hopefully there'll be less opportunity for stylistic errors...

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ This generator can also be further configured with the following command line fl
         --hbs            add handlebars engine support
     -H, --hogan          add hogan.js engine support
     -v, --view <engine>  add view <engine> support (dust|ejs|hbs|hjs|jade|pug|twig|vash) (defaults to jade)
+        --no-view       use HTML directly instead of a view engine
     -c, --css <engine>   add stylesheet <engine> support (less|stylus|compass|sass) (defaults to plain css)
         --git            add .gitignore
     -f, --force          force on non-empty directory
--- a/bin/express-cli.js
+++ b/bin/express-cli.js
@@ -55,6 +55,7 @@ program
   .option('    --hbs', 'add handlebars engine support', renamedOption('--hbs', '--view=hbs'))
   .option('-H, --hogan', 'add hogan.js engine support', renamedOption('--hogan', '--view=hogan'))
   .option('-v, --view <engine>', 'add view <engine> support (dust|ejs|hbs|hjs|jade|pug|twig|vash) (defaults to jade)')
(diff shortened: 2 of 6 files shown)


tobymurray:
Not really sure why the Travis build failed. Any chance it's an intermittent thing? It passes for me locally, and looks suspicious in terms of how it failed - obviously the file is there for other executions of the test (e.g. the appveyor test using the same version of Node).

andrewspiers:
Travis now shows as [passing?](https://travis-ci.org/expressjs/generator/builds/226681347)

tobymurray:
I re-ran the build after the initial failure, it succeeded the second time. Maybe some sort of timing error the first time - I don't believe it has to do with the changes made in this PR.

README.md
```diff
@@ -47,6 +47,7 @@ This generator can also be further configured with the following command line fl
         --pug           add pug engine support
     -H, --hogan         add hogan.js engine support
     -v, --view <engine> add view <engine> support (dust|ejs|hbs|hjs|jade|pug|twig|vash) (defaults to jade)
+        --no-view       use HTML directly instead of a view engine
```
gragonmau: I wonder if it wouldn't be useful to mention API services as well. Something like this:
```--no-view       use HTML directly instead of a view engine, or either build an API service```

tobymurray: I'll leave that up to @dougwilson's discretion, I'm happy with whatever help text makes sense.

JamesBream:
@dougwilson Is this the only issue holding up the 4.16 release?

dougwilson:
There were still some issues I needed to address, but at this point Express 4.17 is going to come out before the 4.16 generator 😆 I'm going to move all the current PRs that are pointing to 4.16 to point to `master` so we can cut a 4.16 generator, since 4.17 is coming very shortly anyway.

dougwilson:
Nice, I'm actually going to land this in 4.16 after all. The branch I have with this rebased and tweaked has everything working, just had to finish off the 404/error handling. Without a templating engine, it cannot contain the specific error message, but it can be better than this PR (which makes even 404s say "An error has occurred" which doesn't make sense).

PR #161, opened 2017-06-13
changed handlebars package

alexanderankin:
[This other package](https://www.npmjs.com/package/express-handlebars) is, for one:

> A Handlebars view engine for Express which doesn't suck.

and also, one which [allows for partials](https://github.com/ericf/express-handlebars) (search for partialsdir on page) which is super useful for me. I'm happy to work with you guys to figure out how to get this integrated into this package, cause theres a handful of live sites/private apis that i made from this tool and its super useful for me, and now that I'm getting into web development more commercially, i found myself editing this bit in and figured its time to give back.

### DIFF
--- a/bin/express-cli.js
+++ b/bin/express-cli.js
@@ -257,6 +257,18 @@ function createApplication (name, path) {
           render: 'adaro.dust()'
         }
         break
+      case 'hbs':
+        app.locals.modules.handlebars = 'express-handlebars'
+        app.locals.view = {
+          engine: program.view,
+          render: 'handlebars.create({\n'
+            + '  extname:       \'hbs\',\n'
+            + '  layoutsDir:    path.join(__dirname, \'views\'),\n'
+            + '  partialsDir:   path.join(__dirname, \'views\', \'partials\'),\n'
+            + '  defaultLayout: \'layout\'\n'
+            + '}).engine'
+        }
+        break
       default:
         app.locals.view = {
           engine: program.view
@@ -296,7 +308,7 @@ function createApplication (name, path) {
         pkg.dependencies['hjs'] = '~0.0.6'
         break
       case 'hbs':
-        pkg.dependencies['hbs'] = '~4.0.1'
+        pkg.dependencies['express-handlebars'] = '^3.0.0'
         break
       case 'pug':
         pkg.dependencies['pug'] = '~2.0.0-beta11'
--- a/bin/express-cli.js
+++ b/bin/express-cli.js
@@ -308,7 +308,7 @@ function createApplication (name, path) {
         pkg.dependencies['hjs'] = '~0.0.6'
         break
       case 'hbs':
-        pkg.dependencies['express-handlebars'] = '^3.0.0'
+        pkg.dependencies['express-handlebars'] = '~3.0.0'
         break
       case 'pug':
         pkg.dependencies['pug'] = '~2.0.0-beta11'
--- a/bin/express-cli.js
+++ b/bin/express-cli.js
@@ -261,12 +261,12 @@ function createApplication (name, path) {
         app.locals.modules.handlebars = 'express-handlebars'
         app.locals.view = {
           engine: program.view,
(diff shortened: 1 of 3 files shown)


bin/express-cli.js
```diff
@@ -296,7 +308,7 @@ function createApplication (name, path) {
         pkg.dependencies['hjs'] = '~0.0.6'
         break
       case 'hbs':
-        pkg.dependencies['hbs'] = '~4.0.1'
+        pkg.dependencies['express-handlebars'] = '^3.0.0'
```
dougwilson: Don't use the `^`.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "express-generator",
   "description": "Express' application generator",
-  "version": "4.15.0",
+  "version": "4.16.0",
```
dougwilson: There is no Express.js 4.16.0; this version matches the version of Express. Just remove the version bump from the PR.

alexanderankin: TIL

alexanderankin:
What do you mean by donated to the organization? Its all free open source hehe, amiright?

But seriously, I can take this up with the maintainers of express-handlebars i guess - but I am not sure as to what you mean by donated. I thought that would mean on github, having the repo be under [github.com/expressjs](github.com/expressjs) organization, but hbs seems to have a home at [https://github.com/pillarjs/hbs](https://github.com/pillarjs/hbs).

Also i took a look at hbs package and it seems like a trivial wrapper which is a different take on what the similarly trivial wrapper in express-handlebars does differently. To that end, I dont see why the same work should be repeated for the purpose of attributing the contributions to someone else (this is one of those big reasons why people do oss in the first place?). As it stands, its my path of least resistance to use my patched generator (or paste it in myself everytime) and urge you to reconsider the 'only use expressjs's own', and for you to let it sit, so ill let it sit.

A word about the tests that dont pass: ill pull down changes that fix the part where the beta pug package you guys depend on stops breaking the node@0.10.x build, with '[SyntaxError: Use of const in strict mode.](https://travis-ci.org/expressjs/generator/jobs/243076547#L323)'. (I can also try to fix this one too, but you might know something about issues unrelated to my PR because youre more familiar with the project?)

dougwilson:
Hi @alexanderankin I'm sorry you disagree.

dougwilson:
The whole debacle with jade, then becoming pug, and having it cause our test to fail is exactly the root reason we want to use our own modules as the first class ones here other third party ones. Third parties seems to keep burning us over and over.

dougwilson:
Perhaps an alternative is to add this module, express-handlebars as an additional option instead of replacing the hbs module?

alexanderankin:
yeah that sounds like good middle ground, what should the options tag read? i could see it being "--hbs" vs "--handlebars"?

dougwilson:
So for the far future, with trying to test everything like we do now, seems to be hitting a wall. I want to at some point pull the generator apart into pieces, pieces that can just be plugged into. I know that doesn't directly relate into, but sort of.

Right now the --hbs is an old, deprecated option (I believe using it even prints a warning), with the current hbs being --view=hbs. I know not a lot of flexibility has been added there, but I think like --view=handlebars may be good enough (vs --view=hbs). Other than that, we could claim that there is a separate --view-options to control various options for the selected view, like picking the underlying engine. Maybe --view=hbs --view-options engine=express-handlebars ?? this seems wordy and not sure if that works with our arg parser. Perhaps even --view=hbs --view-module=express-handlebars ??

I wonder what others think. I can re-open the PR for this discussion, but having a discussion in a pull request is not always obvious, especially since the GitHub workflow is issue = an issue (like should use express-handlebars or something) and a pr = a proposed implementation for the issue. Perhaps we open a new issue to consolidate this discussion + pr + whatever else for everyone to get their heads around in a single place.

alexanderankin:
do you want to create a new issue or should I? I'll draft up something that sounds like what you wrote meanwhile since it looks like this is what im coding on tonight anyways...

PR #168, opened 2017-08-10
Add `.DS_Store` to gitignore

joewagner:
Thought I would start a discussion around this.  I don't have a strong opinion either way, but I'm always manually adding this to `.gitignore` since most projects I work on have at least one developer on a Mac.

### DIFF
--- a/templates/js/gitignore
+++ b/templates/js/gitignore
@@ -28,3 +28,7 @@ node_modules
 
 # Debug log from npm
 npm-debug.log
+
+# File system
+.DS_Store
+


dougwilson:
You should add it to your global ignores so it applies to all projects. Otherwise, adding this means we should add a rule for all special files for every OS, since you never know what OS folks will use.

dougwilson:
https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

joewagner:
@dougwilson That seems reasonable.  I'll close this

PR #180, opened 2017-11-17
update to node-sass 4.x

gasp:
node-sass-middleware 0.11.x updates to node-sass ^4.5.0
node-sass 3.x builds may fail with modern OSX versions v4.x.x versions are more stable


dougwilson:
I don't think we have automated tests for sass in our suite currently, so didn't have an automatic way to find this answer: what the the minimum Node.js version required for this new version of sass?

dougwilson:
I was hoping that pushing the commit to the 4.16 branch would automatically close this pull request, but I guess not. The version support looks good, so merging this in now 👍

PR #183, opened 2017-11-25
Make pug the default template engine instead of jade

YasharF:
Make pug the default template engine instead of jade.  Jade has been renamed to pug due to trademark issues.  The last version of jade has known vulnerabilities per https://snyk.io/test/npm/jade which results in express-generator projects by default have vulnerabilities in their dependencies out of the box.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ This generator can also be further configured with the following command line fl
         --pug            add pug engine support
         --hbs            add handlebars engine support
     -H, --hogan          add hogan.js engine support
-    -v, --view <engine>  add view <engine> support (dust|ejs|hbs|hjs|jade|pug|twig|vash) (defaults to jade)
+    -v, --view <engine>  add view <engine> support (dust|ejs|hbs|hjs|jade|pug|twig|vash) (defaults to pug)
     -c, --css <engine>   add stylesheet <engine> support (less|stylus|compass|sass) (defaults to plain css)
         --git            add .gitignore
     -f, --force          force on non-empty directory
--- a/bin/express-cli.js
+++ b/bin/express-cli.js
@@ -52,7 +52,7 @@ program
   .option('    --pug', 'add pug engine support', renamedOption('--pug', '--view=pug'))
   .option('    --hbs', 'add handlebars engine support', renamedOption('--hbs', '--view=hbs'))
   .option('-H, --hogan', 'add hogan.js engine support', renamedOption('--hogan', '--view=hogan'))
-  .option('-v, --view <engine>', 'add view <engine> support (dust|ejs|hbs|hjs|jade|pug|twig|vash) (defaults to jade)')
+  .option('-v, --view <engine>', 'add view <engine> support (dust|ejs|hbs|hjs|jade|pug|twig|vash) (defaults to pug)')
   .option('-c, --css <engine>', 'add stylesheet <engine> support (less|stylus|compass|sass) (defaults to plain css)')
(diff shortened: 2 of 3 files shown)


YasharF:
@dougwilson either the support policy in express-generator or express itself needs to change to drop support for versions of NodeJS that are in End-of-Life in order to follow standard security practices.  Do you think if I should open an issue with express-generator for the discussion, or would you rather keep the NodeJS version support in express's court (like in https://github.com/expressjs/express/issues/2755 )

dougwilson:
We are dropping support for older version in Express 5 (there is already an open issue and has a recorded discussion people can listen to). Express 5 is still aways out, though. Do you want to resolve this jade issue now in 4.x or wait unt 5?

Also just noticed pug is stll a release candidate.  Please use a stable version for the default engine.

dougwilson:
Closing because these changes are against what we talked about in the TSC and voted for. We welcome your contributions, but if you are looking to help, the best would be to implement the didection that was voted for (it's right at the top of the issue).

YasharF:
From the perspective of security vulnerability management it needs to be resolved in 4.x, since 5.x is ways out.  In other words, it should get addressed in the current stable release of express-generator, or whatever `npm install -g express-generator` ends up deploying.

Regarding the TCS discussion and voting, I think End-of-Life and vulnerability management practices need to get a slightly more weight in the future discussions.  End-of-Life (EOL) software by definition is considered as vulnerable. I'm not sure what the background of current TSC members in vulnerability management is, and what the proper method for the process change might be.

So would the following be the right set of action items?

1. Contacting pug maintainers and having them release 2.0.0 as a stable version (currently it's rc4) - I'll take the action item under https://github.com/pugjs/pug/issues/2610
2. Contacting Node Security Project and snyk.io and ask them to mark express-generator 4.x npm package with a vulnerability, since by default express-generator is creating projects that have vulnerable dependencies - I can take care of this too.
3. Patch express-generator 4.x to change the default to address NSP/snyk flags, this may mean that 4.17 or may be a different 4.x version is going to draw a line in the sand for the default setups using express-generator when it comes to NodeJS<4.x.  It needs to be documented that users on EOL versions of Node will need to explicitly specify the jade flag and need to be aware of the risk of vulnerabilities. - action item dougwilson?
4. Discuss dependency and platform End-of-Life and vulnerability management with TCS voting members, and perhaps engage other folks from NSP or Node.js Foundation as needed. - action item dougwilson?

dougwilson:
The default engine MUST support Node.js 0.10 and up, there is no discussion on this point. Just choose a default engine with that support. If you disagree, did you listen to our TSC discussion on this? If pug is not goinf to support Node.js 0.10 or release a stable version, then we are not goig to change to pug, which was the discussion and why we are going to default to _no view engine_. Defaulting to no engine satisfies your security argument, does it not? If it doesn't please explain.

dougwilson:
The reason we are in this situation is because this is a dependency we cannot control the lifecycle and other aspects. We are not going to keep making those same mistakes. We will not default to a ciew engine we so not have direct control over if it doesn't follow our lifecycles. We are effectively held hostage by our dependency where it has a securuty vulnerability and there isn't even a stable version of it without a vulnerability.  My only action plan is to implement the no view engine. Also for nos I'm blocking yoh from the org and erasing this discussion as a violation of our security policy for public disclosure of vulnerabilities.

It also seems very rude if you are security conscious to publicly disclose this without even a chance to work though a private disclosure and release timeline over a holiday weekend no less (you're in the same country it seems).

PR #185, opened 2017-12-07
tests: check sass

gasp:
having sass tested will help check node compatibility for PR #180

### DIFF
--- a/test/cmd.js
+++ b/test/cmd.js
@@ -316,6 +316,62 @@ describe('express(1)', function () {
       })
     })
 
+    describe('sass', function () {
+      var ctx = setupTestEnvironment(this.fullTitle())
+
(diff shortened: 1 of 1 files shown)


dougwilson:
Hi @gasp thanks for trying to add some tests for `sass`! I know I tried a long time ago but never got them to work. I did rewrite how the testing is done since then and didn't try `sass` again. It does seem like the changes here don't work at all (`npm test` fails on all platforms), but it may be a minor error just the file count on disk is not the same as the count you're expecting to be there. Do you know which file would be missing?

gasp:
hi @dougwilson, I will have a look :-)

dougwilson:
So I ran it locally and it looks like just changing the file count. You don't need to change anything at this point, I'll merge your changes with the modifications I have made for you!

PR #187, opened 2018-01-05
Add JSX as template language

Ayc0:


bin/express-cli.js
```diff
@@ -51,8 +51,9 @@ program
   .option('-e, --ejs', 'add ejs engine support', renamedOption('--ejs', '--view=ejs'))
   .option('    --pug', 'add pug engine support', renamedOption('--pug', '--view=pug'))
   .option('    --hbs', 'add handlebars engine support', renamedOption('--hbs', '--view=hbs'))
+  .option('    --jsx', 'add react jsx engine support', renamedOption('--jsx', '--view=jsx'))
```
dougwilson: You don't want to ad a new option that is immediately deprecated :) You can remove this `--jsx` option 👍

bin/express-cli.js
```diff
@@ -51,8 +51,9 @@ program
...
   .option('-H, --hogan', 'add hogan.js engine support', renamedOption('--hogan', '--view=hogan'))
-  .option('-v, --view <engine>', 'add view <engine> support (dust|ejs|hbs|hjs|jade|pug|twig|vash) (defaults to jade)')
+  .option('-v, --view <engine>', 'add view <engine> support (dust|ejs|hbs|hjs|jade|pug|twig|vash|jsx) (defaults to jade)')
```
dougwilson: You can alphabetize the `jsx` with in the list 👍

bin/express-cli.js
```diff
@@ -308,6 +319,11 @@ function createApplication (name, path) {
       case 'vash':
         pkg.dependencies['vash'] = '~0.12.2'
         break
+      case 'jsx':
+        pkg.dependencies['react'] = '^16.2.0'
```
dougwilson: From our current practices, these should at most be `~` or just pinned; no `^` dependencies :)

bin/express-cli.js
```diff
@@ -442,6 +458,7 @@ function main () {
     if (program.hbs) program.view = 'hbs'
     if (program.hogan) program.view = 'hjs'
     if (program.pug) program.view = 'pug'
+    if (program.jsx) program.view = 'jsx'
```
dougwilson: If you move the `if` statement up a couple lines it'll keep the list alphabetical.

dougwilson:
Forgot to add: update the README as well when you get a chance 👍

bin/express-cli.js
Ayc0: the --engine are for deprecated engines ?

dougwilson: I'm not sure what exactly you mean. When you run `express --jsx` in your PR you have it printing a warning saying that `--jsx` should no longer be used and use `--view=jsx` instead. Just don't add it at all, there doesn't seem to be a point.

Ayc0: kk

Ayc0:
@dougwilson I think everything is set now

imcodingideas:
@dougwilson is there anything I can do to help move this PR along?

dougwilson:
Oh, nice, sorry I missed the comment earlier :+1: looks like the CI is failing, though. If someone wants to get that working that would be 👌

Ayc0:
@dougwilson i don't think I can fix the CI: the tests are failing for the old versions of node (node 4 and previous versions)

dougwilson:
Gotcha. We have never had a lang that didn't work on all our supported versions. We should probably think about what the user experience should be here. Should the generator error our if that lang is chosen when running on a Node.js version the lang does not support? I also wonder if the generator should actually add an engines to the generated package in this case as well. Thoughts?

dougwilson:
It should probably be clear in the documentation that the requirements when choosing the lang is different from everything else and what it is.

Ayc0:
I don't understand what you mean

Ayc0:
I don't understand what you mean

Ayc0:
Node dropped its support for pre v4 versions
The v6 is the current LTS version but Node 4 is still maintained until April 2018

dougwilson:
> engines field
> I don't understand what you mean

In the `package.json` file, there is the field that specifies the compatible Node.js versions: https://docs.npmjs.com/files/package.json#engines

> Node dropped its support for pre v4 versions
> The v6 is the current LTS version but Node 4 is still maintained until April 2018

We still support those versions, just the same as many software applications (even things like Chrome and Firefox) supported Windows XP well after Microsoft stopped supporting it. We will drop those versions on the next major version. If you wish to punt on the issue, we're happy to hold onto this PR until the time comes in which a new major version is released, of course 👍 we have no problem with this.

sant123:
What about transpiling to ES5? Should that work?

Ayc0:
node 4 is deprecated now

PR #170, opened 2017-08-26
ES2015 standards support

sisisin:
#167

PR #188, opened 2018-01-22
Upgrading JS files to ES6, mild dependency tweaks

donburks:
With this PR, I am upgrading the syntax of `app.js`, route files, and `bin/www` to ES6, by converting `var`s to `const`s and converting the callback functions from using `function` to fat-arrow functions.

As a separate forward movement for scaffolding Express applications, I am proposing in this PR to include the [helmet](https://www.npmjs.com/package/helmet) NPM module for security. Appropriate changes have been made to the CLI script to accommodate for this package and its versioning.

These changes are ones that I currently have to make every time I use the express generator, and as I know many other developers making these and other changes to modernize syntax and/or best practices, I am proposing this as a baseline improvement.

### DIFF
--- a/bin/express-cli.js
+++ b/bin/express-cli.js
@@ -279,6 +279,7 @@ function createApplication (name, path) {
         'debug': '~2.6.9',
         'express': '~4.15.5',
         'morgan': '~1.9.0',
+        'helmet': '~3.9.0',
         'serve-favicon': '~2.4.5'
       }
     }
--- a/templates/js/app.js.ejs
+++ b/templates/js/app.js.ejs
@@ -1,17 +1,17 @@
-var express = require('express');
-var path = require('path');
-var favicon = require('serve-favicon');
-var logger = require('morgan');
-var cookieParser = require('cookie-parser');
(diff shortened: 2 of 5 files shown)


JamesBream:
Express generator should probably support ES6 syntax I agree, but as an option and not as the default. Express 4.x, as far as I can tell, still supports node v0.10 and up, which means that this change completely breaks support for those old versions.

As for including helmet, it's a nice idea but I think it's best to keep the generator as skeleton-like as possible. If users want to install and use Helmet then it's ridiculously easy to do, so I don't really see the benefit of including it in the generator - though if you use it personally there's nothing stopping you from editing your own installation to include it.

dougwilson:
So there is already a PR open working on making ES6 an option (even perhaps changing the default value depending on the running Node.js version to make it "just work"). So this does seem like a duplicate of PR #170 

As for the `helmet` change, I'm on the fence, and kinda with @JamesBream above. At the very least, feel free to open an issue regarding it (or a PR with just that change) so it can be discussed 👍 if you really want to see it happen.

PR #191, opened 2018-02-10
Remove redundant code

zhuweiyou:
Because `process.exit(1)` has exited the process, `break` is redundant.

If you use eslint, the error is as follows.

image


dougwilson:
Seems to make sense. The generated code, though, doesn't match StandardJS style currently (maybe it should?), so users applying standard would have way more things to fix in addition to removing those two `break` statements.

zhuweiyou:
Thank you for your reply.

Even if no style checks are used, `break` is also an unreachable code.

What do you think ?

PR #197, opened 2018-06-06
When forced, check if there's an existing package.json and clone dependencies

codinger41:
When the command is forced on a directory that isn't empty, my changes will check if the user has an existing package.json file and clone it's dependencies instead of overwriting it.

### DIFF
--- a/bin/express-cli.js
+++ b/bin/express-cli.js
@@ -133,10 +133,20 @@ function copyTemplateMulti (fromDir, toDir, nameGlob) {
  * @param {string} dir
  */
 
-function createApplication (name, dir) {
+function createApplication (name, dir, force) {
   console.log()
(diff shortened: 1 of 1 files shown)


leksyib:
@dougwilson hey man, do you have any idea why the tests aren't passing on on travis? it passes on appveyor and locally.

leksyib:
@rwky

rwky:
Probably a good idea if it merges here too https://github.com/expressjs/generator/blob/master/bin/express-cli.js#L477

Tests need adding and README needs updating.

dougwilson:
I'm going to close this for now and the requested changed have not been made after a year. If you are still interested in making them, feel free to open a new pull request with the requested changes made.

Issue #199, opened 2018-06-17
One of the biggest feature is missing on Express generator.... 😱😱😱


AJ-Creations:
Hi Dev,
     Today, I was going to create an Express based website on my cloud account. I knew that the providers also provides free SSL for my websites. So, after deployment, I tried to visit my website using https(https://my domain.tld) but the server says that they can't provide secure connection and returns invalid response. Because the server is actually started for http protocol. When I searched about it on Google, everyone says that I have to edit the www file using a text editors manually. If I have to do things manually, then what is the job of Express generator ?

If the Dev team could add a flag like --https then we could use something like Express --view=ejs --https and it'll be more awesome!
Please do it....

dougwilson:
Pull requests welcome to add the feature :+1:

dougwilson:
I'm not sure I understand your question. As for making a pull request, this project is open source and contains what was contributed. You're welcome to contribute the feature you're looking for, but if you're not sure how or have the time, you're welcome to find someone else who has the time / skills to get it added 👍

AJ-Creations:
I'm not that much well learnt.
If anyone could add that awesome feature, I'll describe the thing for them.

Normally, when we start a project by generating Express using command like `express --view=ejs`, It just simply creates an Express app in the folder with ejs chosen as view engine. I'm not talking about EJS okay...
But the problem is that if express-generator users install Express on a cloud computing server (which server a subdomain for him with https enabled),  then he can surely visit the newly created website using address like http://sub.domain.com but not http`s`://sub.domain.com because codes from www.js are creating Web server for only http. Not for https.



If Express generator has another feature like an https flag,
Express-generator user could simply run a command like Express --view=ejs --https for generating Express files with the code of https enabled server.

Please....
If anybody could add this feature, add it. 
Please..

jonnyka:
@dougwilson I've created a PR for the issue, can you check please? thanks

slidenerd:
My suggestion is to NOT add this feature.

- HTTPS is supposed to be handled by a web-server like NGINX
- OR the elastic load balancer
- The application should keep itself free of attaching PEM certificates and stuff

kingadimu:
Your question was answered. Sorta. 
https://help.dreamhost.com/hc/en-us/articles/215747758-Force-your-site-to-load-securely-with-an-htaccess-file

oviecodes:
We could allow the user choose an https option through the command line then the application(express-generator) Could generate an SSL certificate if https option was chosen.

[...]

PR #202, opened 2018-06-19
Add https support

jonnyka:
solution for https://github.com/expressjs/generator/issues/199

### DIFF
--- a/bin/express-cli.js
+++ b/bin/express-cli.js
@@ -9,6 +9,7 @@ var program = require('commander')
 var readline = require('readline')
 var sortedObject = require('sorted-object')
 var util = require('util')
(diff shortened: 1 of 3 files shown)


rwky:
Travis is failing but appears to be due to old node versions, perhaps anything less than 6 should be removed from travis since they're not supported upstream @dougwilson what do you think?

Apart from that @jonnyka can you add tests and update the README?

jonnyka:
@rwky yes will do, I'm a bit swamped today, probably gonna update the PR tomorrow

AJ-Creations:
Awesome!

jonnyka:
@rwky @dougwilson tests are added, readme updated

PR #204, opened 2018-06-20
Adding console.log for yarn

yigitkurtcu:
There is no log support for yarn start or yarn install.I added console.log for yarn.

### DIFF
--- a/bin/express-cli.js
+++ b/bin/express-cli.js
@@ -346,13 +346,14 @@ function createApplication (name, dir) {
   console.log()
   console.log('   install dependencies:')
   console.log('     %s npm install', prompt)
+  console.log('     %s yarn install', prompt)
   console.log()
   console.log('   run the app:')
 
   if (launchedFromCmd()) {
-    console.log('     %s SET DEBUG=%s:* & npm start', prompt, name)
+    console.log('     %s SET DEBUG=%s:* & npm start or yarn start', prompt, name)
   } else {
-    console.log('     %s DEBUG=%s:* npm start', prompt, name)
+    console.log('     %s DEBUG=%s:* npm start or yarn start', prompt, name)
   }
 
   console.log()


bin/express-cli.js
```diff
@@ -346,13 +346,14 @@ function createApplication (name, dir) {
   console.log()
   console.log('   install dependencies:')
   console.log('     %s npm install', prompt)
+  console.log('     %s yarn install', prompt)
   console.log()
   console.log('   run the app:')
 
   if (launchedFromCmd()) {
-    console.log('     %s SET DEBUG=%s:* & npm start', prompt, name)
+    console.log('     %s SET DEBUG=%s:* & npm start or yarn start', prompt, name)
```
dougwilson: This line should be a valid command to copy and paste

bin/express-cli.js
```diff
@@ -346,13 +346,14 @@ function createApplication (name, dir) {
...
   } else {
-    console.log('     %s DEBUG=%s:* npm start', prompt, name)
+    console.log('     %s DEBUG=%s:* npm start or yarn start', prompt, name)
```
dougwilson: Same or this.

yigitkurtcu:
Tests added. It needs to be checked.

dougwilson:
Hi @helio-frota thank you for jumping in to help out here 🥳 . Typically in order to land a PR, we would need the CI to pass. I understand there is something unrelated to the PR preventing it to pass, but we should get that fixed, and then land these. Would you be willing to lend a hand to assist (maybe even PR) in addressing that CI failure spot?

yigitkurtcu:
> LGTM,
> The error on CI seems pretty similar to this other PR #246 which has changes only inside a README.md. So it is failing due another reason.

Thanks for your reviews.

helio-frota:
Hi @dougwilson thank you for the extra information :+1: .
Sounds good, I'll start looking on that CI issue.

helio-frota:
hi @dougwilson PR sent https://github.com/expressjs/generator/pull/265 to fix that CI issue.

Issue #25, opened 2014-05-07
Why did the startup stub get moved to bin/www ?


Todd-Werelius:
Just curious as to the thought behind this, seems a bit odd to make the generated file a module and then call it from a short stub in another directory

PR #205, opened 2018-06-24
Remove "node" from "node ./bin/www"

JBallin:
`./bin/www` is already a node executable so it uses node automatically. https://github.com/expressjs/generator/issues/25 doesn't mention this.

Is there some edge case where `node` is needed?


dougwilson:
Hi! Sorry if that issue is long. I mentioned why the node is necessary in this comment: https://github.com/expressjs/generator/issues/25#issuecomment-43613128

You can see that reflected in the CI build failing now that the node was removed as well

PR #207, opened 2018-07-17
Fixes issues where Node doesn't error if port is in use.

### LINKED ISSUE #206: Server listens on port already in use
If that port is already in use by another application when ```npm start``` is run, the server fails to err and starts up as if the port is free to listen on.

The issue is related to the server configuration for events happening after the listen call is made and the listen call not specifying a binding of 0.0.0.0.

bacrossland:
This fixes #206

### DIFF
--- a/templates/js/www.ejs
+++ b/templates/js/www.ejs
@@ -25,9 +25,9 @@ var server = http.createServer(app);
  * Listen on provided port, on all network interfaces.
  */
 
-server.listen(port);
 server.on('error', onError);
 server.on('listening', onListening);
+server.listen(port, '0.0.0.0');
 
 /**
  * Normalize a port into a number, string, or false.


bacrossland:
Working on a solution now. I'm going to update my issue statement as well to clarify more what is happening and why.

bacrossland:
After a lot of research, I have reverted part of my fix and provided test for it. The issue that I was having is an OS specific issue. If you have IPv6 enabled it will not necessarily block IPv4 for listening. I've added a comment block to explain more what is happening incase other engineers have this issue with their app. 

I've also added in a catch for an error condition that may return on Linux when kernel option ``` ipv6.disable=1``` is used. I did not force listening on IPv4 under that condition and instead opted for alerting the engineer of the issue so they can make a decision based on their environment.


Sources:

ipv6only listen/bind option
https://github.com/nodejs/node/issues/17664

Why listening "::" (all IPv6) also get "0.0.0.0"(all IPv4) listened?
https://github.com/nodejs/node/issues/9390

bacrossland:
Sorry I haven't gotten back to this. I will try to get to this before the end of the month.

PR #200, opened 2018-06-17
Add --es6 option to enable ES6 compliant syntax

danielkhan:
This PR 
- adds a new --es6 flag to enable ES6 (ES2015) syntax
- replaces `const` with `var` if --es6 is set
- adds string templates if --es6 is set
- prefers arrow functions for callbacks if --es6 is set
- makes sure that this feature is only available for Node.js >= 6.0.0
- adds tests for --es6

Reference Link https://github.com/expressjs/generator/issues/167

PR #209, opened 2018-09-18
Changed the 'var' in the app.js into 'const'

mayomi1:
Most people normal edit the 'var' in the app.js and change it to 'const', this small PR solved that issue


dougwilson:
PR #200 is already open with a solution :+1: Thanks for the effort, though!

PR #210, opened 2018-09-27
corrected route user

rajat-np:
Route user was pointing to '/', but it should be '/user'


dougwilson:
There are many ways to design REST routes. The way our example lays it out is as we intend.

PR #211, opened 2018-10-15
JS Typing  Enhancements

cagataycali:
Current ESLint which extends standard is allowed for usage without dot.
When I use generator always doing same stuff for beginning coding.

That stuffs have been fixed;
- [x] var -> const conversation
- [x] ESLint & vulnerable package updates
- [x] Traditional functions to arrow functions
> ES2015 arrow function literals have been [integrated %99 after 6.14.4 nodejs version](https://node.green/#ES2015-functions-arrow-functions) and express-generator users which also expecting scaffold new gen express app, who uses newest version of node. That's the reason I removed older versions from CI process.
- [x] ESLint warnings for deprecated node API's
> Errors threw 120 times when lint proceed.
'assert.equal' was deprecated since v10.0.0. Use 'assert.strictEqual' or 'assert.strict.equal' instead
'assert.notEqual' was deprecated since v10.0.0. Use 'assert.notStrictEqual' or 'assert.strict.notEqual' instead
- [x] Removed unnecessary ["next" phrase usage in sub-routes](https://github.com/cagataycali/generator/commit/a4fb85cd5dd9d2b995d8ffa42b0cabd5858d9d19#diff-01d2c7574633e64e9b6a25881f345d85L5)

The lint and mocha test runs successfully.

```
npm run lint; && npm run test
```

The expectation of this PR is to save 2 minutes of the developer.


dougwilson:
Please reopen when you can add back support for all the Node.js versions you removed. Thanks!

PR #212, opened 2018-10-16
JS Typing  Enhancements

cagataycali:
Details;
https://github.com/expressjs/generator/pull/211


dougwilson:
Hi @cagataycali thanks for the pull request! When you have the chance, please take a look into the CI failures. It looks like the failures are likely due to some changes you introduced that may not work correctly on all the Node.js versions this module supports.

The generator will create a Express.js 4.x scaffold and is designed such that both (a) the generator itself runs on the same Node.js versions Express 4.x support (0.10+) and that the generated app works on those environments as well.

For example, a user who has Node.js 4.x installed should be able to install the generator, generate a simple express app with it, and the app runs on the same Node.js 4.x.

I hope that helps!

cagataycali:
If you accept, we can use babel

dougwilson:
I'm not very familiar with how that works exactly. Can you explain what that means and how it would fix it?

dougwilson:
Closing as I am trying to go though this repository and address everything. Two simple change requests on the PR have gone unanswered for 12 months. If you still want to address them, feel free to open a new PR with the requested changed made.

PR #213, opened 2018-10-24
Remove jade

cByst:
It appears that jade has some security vulnerabilities showing up in npm audit. Unfortunately jade has been deprecated and the project has been renamed as pug so to remediate these vulnerabilities we would need to remove jade from the project and default to pug. This PR does that, removes jade from the templates and and replaces the default view engine with pug. Since pug is just a renamed version of jade that is still being supported the impact is minimal.


dougwilson:
Instead, maybe implement as per voted on in #141 for the change of default and leave jade as an option.

PR #214, opened 2018-10-24
Update mocha and supertest dependencies to fix security vulnerabilities

cByst:
Updated the project dependencies to resolve 2 security vulnerabilities. They both are related to the tests so shouldn't be to big of a deal. 

```
found 5 vulnerabilities (2 low, 1 moderate, 1 high, 1 critical) in 379 scanned packages
  5 vulnerabilities require semver-major dependency updates.


dougwilson:
The new versions of these don't work on the old Node.js versions we support, which is why they are the versions they are.

PR #225, opened 2019-02-12
Update README.md

tortuvshin:


dougwilson:
Hi @tortuvshin thank you for taking the time to create a pull request!

Unfortunately this is removing something that is intended to be there: the `'` is an apostrophe, which changes the word into the possessive form in English. The apostrophe makes the line read `Express' application generator` - as in the application generator which belongs to Express.

If you're curious about the possessive form in English, https://en.wikipedia.org/wiki/English_possessive provide a pretty good overview :) 👍

PR #226, opened 2019-03-15
lint: default template to standard

pickfire:
Time to use standard in generator https://github.com/expressjs/discussions/issues/18

### DIFF
--- a/templates/js/app.js.ejs
+++ b/templates/js/app.js.ejs
@@ -1,51 +1,51 @@
 <% if (view) { -%>
-var createError = require('http-errors');
+var createError = require('http-errors')
 <% } -%>
-var express = require('express');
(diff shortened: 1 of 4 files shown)


dougwilson:
Thanks! Can the CI use standard to validate that the generated apps are valid as well? Especially as the other PRs begin to get merged it would be ideal to know if the generated app is still valid or not if it's going to have a specific style.

pickfire:
@dougwilson I have added test for lint but not sure why it does not fail even if there's an extra semi-colon.

dougwilson:
Probably due to the error displayed in our CI:

> Error: No ESLint configuration found.

You would need to add the eslint configuration for standard for it to validate it against that.

PR #228, opened 2019-04-25
Modify of function callback for arrow function and var for const in v…

dcampos1:
Hello @dougwilson 

I think the express generator is a fantastic tool and I would like to make a small contribution by changing the routes of the callback functions by arrow functions, I know it may be an irrelevant change, but the important thing is that the tool is always improved.


dudemelo:
@dougwilson I think this PR can be closed once the author account was deleted.

dougwilson:
Closing.

PR #230, opened 2019-05-29
Update pug from 2.0.0-beta11 to 2.0.3 to fix vulnerability

alevis:


codingthat (changes requested):
I'm not sure which vulnerability this bump fixes, but bumping to 2.0.4 definitely fixes a vulnerability with subdependency clean-css affecting its versions before 4.1.11.

alevis:
@codingthat thanks for the reply!

dougwilson:
I was about to ask why closed, but see you reopened. I guess just pressed the wrong button as they are right next to each other 🤣

codingthat:
It's failing on only the oldest Node.js versions.  Does `generator` itself need to support those?  (I mean...I can't imagine doing a greenfield project on an unsupported version.)

codingthat:
I've seen some projects cap their dependencies in cases like this. If that seems sensible, it could just be a matter of saying, "Want to use generator with Node.js 0.10?  You need to go back to the last known release that worked with it, express-generator X.Y.Z)" in the readme, and then pruning the CI requirements a bit.

mraible:
+1 for fixing this. In the meantime, `npm audit fix` is a workaround.

PR #233, opened 2019-07-12
update isNaN global to use Number.isNaN method

nisalVD:
refer to this.
https://stackoverflow.com/questions/46677774/eslint-unexpected-use-of-isnan


dougwilson:
Duplicate of #234

PR #236, opened 2019-08-17
Changed default view engine to pug

replacepreg:
npm says that jade package includes some security vulnerabilities which is one of them is critical.
so, I changed the default engine immediately.
I didn't tested the app, since I don't know how to do it.


replacepreg:
Changed default view engine from jade to pug.
Also adjusted tests.

sam-github:
I'd like to see this change as well, having security vulns in the default use case isn't something that should be enouraged for express users (or any project).

@replacepreg this needs additional changes:
1. package.engines.node has to change to `> 0.12`
2. package.version has to change to `5.0.0` (because dropping support for a node version is `semver-major`)
3. .travis.yml needs `0.10` and `0.12` removed

Maintainers - opinions?

https://www.npmjs.com/package/jade has been thorougly deprecated, its not a good default.

PR #242, opened 2019-11-06
Dependencies

vansergen:
Update devDependencies to the latest versions and add `package-lock.json`


package.json
```diff
@@ -51,7 +51,7 @@
     "validate-npm-package-name": "3.0.0"
   },
   "engines": {
-    "node": ">= 0.10"
+    "node": ">=6.17.1"
```
dougwilson: I'm not sure what the goal of this PR is, but wanted to at least intervene early here: the goal of the generator is to support generating express projects. The Node.js versions the generator supports matches the versions that express itself supports.

If there is a desire to drop certain versions, express should drop those first.

dougwilson:
Closing due to no response.

PR #244, opened 2019-12-06
remove unsupported Node.js, bump major, update pug

sam-github:
gen: update pug to ~2.0.4
    
    Fixes:
    - https://npmjs.com/advisories/785
    - Dependency on pug's -beta11

    build: add Node.js 10, 12, 13, remove EOL Node.js
    
    Bumps the package version since dropping support for Node.js versions is
    semver-major.
    
    Anyone who is still generating new express apps for out-of-support
    Node.js versions (unadvisedly) can either modify the generated
    dependencies after generation, or use express-generator 4.x

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "express-generator",
   "description": "Express' application generator",
-  "version": "4.16.1",
+  "version": "5.0.0",
   "author": "TJ Holowaychuk <tj@vision-media.ca>",
   "contributors": [
     "Aaron Heckmann <aaron.heckmann+github@gmail.com>",
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,9 @@
 language: node_js
 node_js:
-  - "0.10"
-  - "0.12"
-  - "1.8"
-  - "2.5"
-  - "3.3"
-  - "4.9"
-  - "5.12"
-  - "6.16"
-  - "7.10"
-  - "8.15"
-  - "9.11"
+  - "8"
+  - "10"
+  - "12"
+  - "13"
 matrix:
   fast_finish: true
 sudo: false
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "express-generator",
   "description": "Express' application generator",
-  "version": "4.16.1",
+  "version": "5.0.0",
```
dougwilson: The generator version matches the generated express version. Make sure to update the related version of Express as well, and you can remove this line from the PR, as npm command would then cause us to release as a diff versions since it bumps what is in the package when making the tag and publishing.

sam-github:
@dougwilson thanks for the comments.

express supports 0.10 and later, on both master and 5.x-alpha:
- https://github.com/expressjs/express/blob/e1b45ebd050b6f06aa38cda5aaf0c21708b0c71e/package.json#L82
- https://github.com/expressjs/express/blob/5.x/package.json#L83

I'm not sure how to approach that. Dropping support for a node version should cause a package major bump, so it wouldn't be able to do this until express@6.x, unless the alpha status allows dropping EOL node support?

I found: https://github.com/expressjs/express/issues/2755#issuecomment-382499774

> When 5.0 is released it will support down to the LTS version at that time.

dougwilson:
It would probably go (1) land a change to some express (5, 6, whatever) that removes various Node.js versions from its matrix. Then this module would change its matrix to match that of the express version it'll drop into the package.json it produced.

sam-github:
> It would probably go (1) land a change to some express (5, 6, whatever) that removes various Node.js versions from its matrix.

That PR exists: https://github.com/expressjs/express/pull/3948

So, I should wait until it lands, then refresh this PR?

dougwilson:
Yea, if the goal of this PR is to drop Node.js versions, it should wait.

sam-github:
The goal is for generated apps to not have audit warnings, but as long as CI requires support back to 0.10, and at least one of the deps (pug) doesn't support 0.10 anymore, those two goals are not reconcilable. I'm sure deps will be updated as a matter of course, sometime, so I'll close this. Users can run `npm audit fix` in the meantime.

Issue #62, opened 2015-01-26
add example generated directory.


chirag04:
What do you guys think about adding an example directory which is the generated directory from this generator?

This will serve as a reference to what is going on in the new versions of this generator. And i generally find it handy to know the structure before running the generator when i start a new project.

dougwilson:
Yea, perhaps. I definitely think it's useful to see the generated stuff just online.

nicolasmccurdy:
Personally, I think it would be cool to just have the directory structure in the readme (similar to [tree](http://mama.indstate.edu/users/ice/tree/)'s output. However, I think also having the generated source somewhere might also be useful.

crashspringfield:
I found this discussion while looking for an example of bin/www because it doesn't always show up like it should, but sometimes does (for reasons I can't explain). So on that note, I think having the directory structure somewhere, with the code available too, would be super useful for those of us who are new and still figuring stuff out.

PR #246, opened 2020-02-01
Default project structure tree

jonchurch:
Adds default project structure to readme

Open to suggestions here. Definitely I think that there should be some explanation about the structure, but as it stands just adding this to the readme (IMO) improves the project because it stands to document things as they currently are.

This is in reference to https://github.com/expressjs/generator/issues/62, where I think we can continue to talk about documenting the structure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,31 @@ Start your Express.js app at `http://localhost:3000/`:
 $ npm start
 ```
 
+## Project Structure
+
+The generator creates the following project structure by default:
+
+```
+.
+├── app.js
+├── bin
+│   └── www
+├── package-lock.json
+├── package.json
+├── public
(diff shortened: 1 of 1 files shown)


jonchurch:
I'm sorry, this is off topic but @dougwilson HOW did you label this seconds after I submitted?? Do you have automation setup to do that?

dougwilson:
I interact with GitHub through email and I see & read just about every message that comes in as it comes in :) Depending on what I'm doing at the time I may or may not interact with it, depending on how much time it would take. I manually labeled it when I saw the email and was planning to take a closer look later, haha :)

helio-frota:
The CI error is not related to the update on README.md. It seems to be a flakey travis build.

PR #248, opened 2020-02-05
Add typescript support

louisyuenll:
I want to add TypeScript language support in the generator so that developers can enjoy strong-typed programming and use ES6 features in their applications.


bin/express-cli.js
```diff
@@ -147,12 +148,32 @@ function createApplication (name, dir) {
...
+    devDependencies: {
     }
   }
 
+  if (program.typescript) {
+    pkg.scripts['build'] = 'tsc'
+
+    pkg.devDependencies['typescript'] = '~3.7.5'
+    pkg.devDependencies['@types/node'] = '~13.7.0'
+    pkg.devDependencies['@types/debug'] = '~4.1.5'
```
dougwilson: Should the types package version pach the debug package version?

bin/express-cli.js
```diff
@@ -147,12 +148,32 @@ function createApplication (name, dir) {
...
+    pkg.devDependencies['@types/express'] = '~4.17.2'
```
dougwilson: Should the types package version pach the express package version?

bin/express-cli.js
```diff
@@ -147,12 +148,32 @@ function createApplication (name, dir) {
...
+    },
+    devDependencies: {
     }
   }
 
+  if (program.typescript) {
+    pkg.scripts['build'] = 'tsc'
+
+    pkg.devDependencies['typescript'] = '~3.7.5'
+    pkg.devDependencies['@types/node'] = '~13.7.0'
```
dougwilson: Would this cause an issue when generated and used on an older node.js like, say 10.x?

bin/express-cli.js
```diff
@@ -325,15 +376,27 @@ function createApplication (name, dir) {
...
   // sort dependencies like npm(1)
   pkg.dependencies = sortedObject(pkg.dependencies)
 
   // write files
-  write(path.join(dir, 'app.js'), app.render())
+  var appScriptName
+  var wwwScriptName
+  if (program.typescript) {
+    appScriptName = 'app.ts'
+    wwwScriptName = 'bin/www.ts'
```
dougwilson: The `www` script shouldn't have an extension; it follows the pattern of being an executable (with shebang on the top).

test/cmd.js
```diff
@@ -74,7 +74,8 @@ describe('express(1)', function () {
         '    "http-errors": "~1.6.3",\n' +
         '    "jade": "~1.11.0",\n' +
         '    "morgan": "~1.9.1"\n' +
-        '  }\n' +
+        '  },\n' +
+        '  "devDependencies": {}\n' +
```
dougwilson: It is probably not add this if there are none; `npm` will end up removing it on the next `npm I` command a user uses. That's why we sort them in the same way npm does as well, so it doesn't cause churn in the file after the fact.

dougwilson:
Awesome, I love this idea! I added a few questions/comments above. Note that I'm not familiar with TypeScript, so take them with a gain of salt :) It would be great if a test could be added as well.

bin/express-cli.js
louisyuenll: It should be safe on older node environment as the version of the generated codes are controlled via tsconfig.json (line 5, "target": "es5"). Currently it is set compiling codes down to ES5 to support most of the older node.js.

dougwilson: Gotcha. I guess i mean typescript-wise. As in I assume that the node 13 types include definitions that do not apply/work on node 10, which then reduces the useful type checking that would prevent that?

bin/express-cli.js
louisyuenll: Got it, may be adding a post-build script to create a symbolic link with name www, pointing to www .js?

As some people may want to generate source-maps for the compiled JS files, I afraid directly renaming it may break the file linkage.

For example,

```js
// add postbuild.js

var fs = require('fs')
var path = require('path')

// Create a symlink, linking www to www.js
var wwwScriptPath = path.join(__dirname, 'bin/www.js')
var wwwPath = path.join(__dirname, 'bin/www')
if (fs.existsSync(wwwScriptPath) && !fs.existsSync(wwwPath)) {
  fs.symlinkSync(wwwScriptPath, wwwPath)
}
```

dougwilson: Unfortunately normal users cannot create symbolic links on windows.

louisyuenll: How about if coping www .js as www?

```js
var fs = require('fs')
var path = require('path')

// Polyfill copyFileSync for node < 8.5.0
if (!fs.copyFileSync) {
  fs.copyFileSync = function (src, dest, flags) {
    fs.writeFileSync(dest, fs.readFileSync(src), flags)
  }
}

// Copy www.js to www
var wwwScriptPath = path.join(__dirname, 'bin/www.js')
var wwwPath = path.join(__dirname, 'bin/www')
if (fs.existsSync(wwwScriptPath)) {
  fs.copyFileSync(wwwScriptPath, wwwPath)
}
```

bin/express-cli.js
louisyuenll: Yes, you are correct.

As the typing version of node is not one-to-one, perhaps we can adjust it to match the major version of current node.

However, during testing I found that node typing below v8 causes typing mismatch errors so that I suggest setting the minimum of @types/node to v8.

```
// When below v8 (tested with v6.17.1):

node_modules/@types/connect/index.d.ts:21:42 - error TS2689: Cannot extend an interface 'http.IncomingMessage'. Did you mean 'implements'?

21     export class IncomingMessage extends http.IncomingMessage {
```

```js
pkg.devDependencies['@types/node'] = '^' + Math.max(8, parseInt(process.version.replace('v', '')))
// example output => "@types/node": "^12"
```

bin/express-cli.js
louisyuenll: Agree, better align with express v4.16.1.
Per check, @types/express also support typing on v4.16.

test/cmd.js
louisyuenll: Got it, will remove it and only add devDependencies on TypeScript templates.

```js
pkg.devDependencies = {}
pkg.devDependencies['typescript'] = '~3.7.5'
// ...rest of codes
```

bin/express-cli.js
```diff
@@ -158,7 +158,9 @@ function createApplication (name, dir) {
     pkg.devDependencies['typescript'] = '~3.7.5'
     pkg.devDependencies['@types/node'] = '~13.7.0'
     pkg.devDependencies['@types/debug'] = '~4.1.5'
-    pkg.devDependencies['@types/express'] = '~4.17.2'
+    pkg.devDependencies['@types/express'] = '~4.16.1'
+
+    pkg.dependencies['tslib'] = '~1.10.0'
```
louisyuenll: As current settings automatically add the helper codes (e.g. ES6+ polyfills) into every related output files, I added [tslib](https://github.com/microsoft/tslib) so that code duplication can be avoided.

samsan1212:
Hi  @dougwilson ,
May I know if this feature is released?
I also love Typescript. It would be so nice if express generator supports typescript.
Is there anything I can help?

validity-check:
When will this be released? I can't stand using plain javascript

prince-t-asamoah:
> When will this be released? I can't stand using plain javascript

Same here, looks like nothing been done since 4 years

PR #250, opened 2020-02-06
should use a 501 not a 404

ORESoftware:
should use a 501 not a 404

### DIFF
--- a/templates/js/app.js.ejs
+++ b/templates/js/app.js.ejs
@@ -31,9 +31,9 @@ app.use(<%= mount.path %>, <%- mount.code %>);
 <% }); -%>
 
 <% if (view) { -%>
-// catch 404 and forward to error handler
+// catch 501 and forward to error handler
 app.use(function(req, res, next) {
-  next(createError(404));
+  next(createError(501));
 });
 
 // error handler


dougwilson:
You're welcome to do this in your own app, but I don't agree for this to be the default.

dougwilson:
To follow up with the reasoning here: the related issue linked to https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501 for a 501 description, in which it states the following:

> 501 is the appropriate response when the server does not recognize the request method and is incapable of supporting it for any resource. The only methods that servers are required to support (and therefore that must not return 501) are GET and HEAD.
> 
> If the server does recognize the method, but intentionally does not support it, the appropriate response is 405 Method Not Allowed.

Node.js already rejects unknown HTTP methods (since it is the server) and thus whatever gets down to Express are methods that Node.js does indeed support.

The article clarifies that GET and HEAD must not return a 501, since the servers must support those methods, though this PR will return a 501 for those two methods.

The second paragraph says that is the server does recognize the method, but does not support it intentionally, it should return a 405. Since Node.js already rejects unsupported methods, that would leave 405 to be the only one here in the Express realm, though the user would need to choose to not support that method intentionally (Express supports all the methods by default that are known by Node.js).

ORESoftware:
Yeah but the generator is not for the "Express realm", it's for end users APIs, right? 405 or 501 seems like a good default for most people's applications.

dougwilson:
Hmm, I'm not sure what you mean by the "Express realm", but my reasoning is explained above, confirmed by the MDN docs you linked to, that it's not the right code to use with the way it was implemented in the PR.

Remember, MDN says that GET and HEAD methods can never return a 501, but this PR does that anyway.

ORESoftware:
The RFC standards aren't perfect, but I guess we have to go by them, so then 405 makes more sense than 501.

PR #252, opened 2020-02-27
Updated JS files to ES6 syntax

georgie105:
I have made some modifications on variable declarations that were written with "var" to "const".
I also modified the functions to use the "arrow function" syntax.


danielkhan:
Duplicate to
- https://github.com/expressjs/generator/pull/200
which unfortunately never merged.

import-brain (approved):

import-brain (changes requested):
There's a lot of CI issues here

PR #256, opened 2020-03-21
Upgraded outdated dependencies in template

KoolTheba:
Hi @dougwilson !

I've just installed the express-generator and I saw that the dependencies were outdated. Please let me know if this modification is ok for you.

**Dependencies upgraded:**

```
Package      Current  Wanted  Latest  Location
debug          2.6.9   2.6.9   4.1.1  api
express       4.16.4  4.16.4  4.17.1  api
http-errors    1.6.3   1.6.3   1.7.3  api
morgan         1.9.1   1.9.1  1.10.0  api
```

### DIFF
--- a/bin/express-cli.js
+++ b/bin/express-cli.js
@@ -145,8 +145,8 @@ function createApplication (name, dir) {
       start: 'node ./bin/www'
     },
     dependencies: {
-      'debug': '~2.6.9',
-      'express': '~4.16.1'
+      'debug': '~4.1.1',
+      'express': '~4.17.1'
     }
   }
 
@@ -166,7 +166,7 @@ function createApplication (name, dir) {
   // Request logger
   app.locals.modules.logger = 'morgan'
   app.locals.uses.push("logger('dev')")
-  pkg.dependencies.morgan = '~1.9.1'
+  pkg.dependencies.morgan = '~1.10.0'
 
   // Body parsers
   app.locals.uses.push('express.json()')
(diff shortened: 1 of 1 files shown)


dougwilson:
Thanks my friend! Looks like we just need to update the unit tests to account for the version changes. Perhaps we should make the unit tests ignore the specific version, as this keeps coming up 🤔

helio-frota:
@dougwilson makes sense :+1: 

The test is to check if it has a package.json file, then the original version of the dependencies (modified in express-cli.js) can be added as part of the content, something like:

```js
it('should have a package.json file', function () {
      var file = path.resolve(ctx.dir, 'package.json')
      var fileLoaded = require(file)  // <--- Changed here.
      var contents = fs.readFileSync(file, 'utf8')
      assert.strictEqual(contents, '{\n' +
        '  "name": "express-1-no-args",\n' +
        '  "version": "0.0.0",\n' +
        '  "private": true,\n' +
        '  "scripts": {\n' +
        '    "start": "node ./bin/www"\n' +
        '  },\n' +
        '  "dependencies": {\n' +
        '    "cookie-parser": "~1.4.4",\n' +
        '    "debug": "' + fileLoaded.dependencies.debug + '",\n' +  // <--- Changed here.
        '    "express": "~4.16.1",\n' +
        '    "http-errors": "~1.6.3",\n' +
        '    "jade": "~1.11.0",\n' +
        '    "morgan": "~1.9.1"\n' +
        '  }\n' +
        '}\n')
    })
```

heardofdezz:
Thanks for the update

PR #265, opened 2020-06-21
build: fix issue on CI related to node-sass

helio-frota:
This commit changes Node.js to a specific 1.8.x version to help CI be
able to find a correct native node-sass binding.

Also the maxBuffer of the child_process exec function was increased
to hold the compilation of the native binding.


dougwilson:
Hello, thanks for the fix. Sorry it sat for so long, and I am trying to go through to address issues here. I wasn't able to merge due to the fork being deleted, but the buffer fix was landed with adjustments to make it a variable and such.

PR #269, opened 2020-07-21
Adds SIGTERM listener to gracefully shutdown the server

HarryEMartland:
Kubernetes and other process managers send a SIGTERM to applications to give them time to shut down. This change adds a listener to this event which shuts the server down gracefully.

### DIFF
--- a/templates/js/www.ejs
+++ b/templates/js/www.ejs
@@ -88,3 +88,10 @@ function onListening() {
     : 'port ' + addr.port;
   debug('Listening on ' + bind);
 }
+
+process.on('SIGTERM', () => {
+    debug('SIGTERM signal received: closing server');
+    server.close(() => {
+        debug('Http server closed');
+    });
+});


HarryEMartland:
@dougwilson sorry I must have missed a notification adding the tags. When you say needs tests what kind of tests would you be looking for? Thanks

UlisesGascon:
maybe @dougwilson can re-run the tests for `Job #792.3`. Seems like there was a error on `node-sass` building process (not relevant to the PR change) [Direct Link](https://travis-ci.org/github/expressjs/generator/jobs/710537959)

HarryEMartland:
If someone could give me some pointers on how to test this I would greatly appreciate it. The issue is only really noticeable if a request takes a decent amount of time. 

In sudo code I would want to test something like this;

```
Start app
Make a request to a slow endpoint
Send SIGTERM to process
Wait for the request to finish
Verify the request returned successfully
Verify the process exited
```

I have had a play with the existing tests and can see that there is code already sending SIGTERM. I am however unsure of how to add a new endpoint with a delay in it.

PR #270, opened 2020-07-24
added `localhost` option for hostname.

RishabhKodes:


dougwilson:
Thank you for this, but the JavaScript expression `'127.0.0.1' || 'localhost'` will never return the second value added here, as the first value is a fixed string and will always be true.

RishabhKodes:
> Thank you for this, but the JavaScript expression `'127.0.0.1' || 'localhost'` will never return the second value added here, as the first value is a fixed string and will always be true.

Yes, you are right, thanks for the quick response.

PR #278, opened 2020-08-14
Showing the running port in console logs.

BashCloud:


dougwilson:
Hello! The port will already be displayed when running the server. Just follow the command shown on how to start the server at the end of the prompt when you generate your project. The message you will see when following along will also not print until the server is actually listening on that port, whereas this change will display the message before Express is actually listening.

PR #280, opened 2020-10-01
Convert to ES2015

redaktice:
Update templates to use `const` and arrow functions.
Update test JS files to use ES2015 syntax

Inspired by https://github.com/expressjs/generator/pull/252


gfeller:
Have been waiting a long time for this change. 
Any chance that this pull request gets merged?

dougwilson:
Hi @gfeller , can you ellaborate on the goal of this PR? It seems to change the style of both the internal generator code and the templated code. Is this desired?

There is nothing wrong with adding a newer es templated code that is an option to the generator, of course, though it doesn't seem like this PR keeps the old format around still under such an option.

The actual generator code it seems to no longer follow the StandardJS format, which we require our org projects to follow, like this one was doing before. Does StandardJS prevent the change to the new style or something?

dougwilson (changes requested):
Internal generator code must follow StandardJS style.

redaktice:
> Internal generator code must follow StandardJS style.

No problem @dougwilson, I made the change back to the StandardJS format

PR #281, opened 2020-10-03
Adding JSX (React) as an option for the view engine in express

Reshav-Paul:
The Express official docs list react as an option for the view engine using the express-react-views module.
This pull request adds jsx as an option for the view engine from the CLI using --view = jsx.

Commits have been made for -

1. Adding CLI support for jsx.
2. Default jsx templates
3. Tests to validate its working.
4. Updated README to include jsx

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ This generator can also be further configured with the following command line fl
         --pug            add pug engine support
         --hbs            add handlebars engine support
     -H, --hogan          add hogan.js engine support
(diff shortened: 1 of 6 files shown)


Reshav-Paul:
Builds are failing on older versions of nodejs. It is passing fine otherwise.

dougwilson:
Thank you! Do not worry about the failing older Node.js versions for this, as it does pass fine on the newer versions. I am adding a method to target a given Node.js version for the upcoming version, so this template would just fit in by requiring a minimum target version.

Reshav-Paul:
That would be great! Thanks for the update.

LucaSorvillo:
why express doesnt add jsx support?

PR #286, opened 2020-12-27
feat: build and run from Docker Support

Rajpratik71:
Build and run the generator from docker image support added.

The main aim and motivation behind doing this is, because
of a no. of different generator are there,
which are compatible and works with the different version of generator
and also might be incompatible with other generator or versions of generator

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>


dougwilson:
Thnk you. There is certainly a lot of things going on in that file and we should add a test for this to the CI, so we can better maintain it.

PR #294, opened 2021-03-26
update dir naming

gumbol:
hi first of all great app.
I changed the dir naming from javascript and stylesheets to scripts and styles.
Scripts because there are non js scripts out there.
styles for short and because (from what i have looked up on the internet) its consistent with many web directory structures

### DIFF
--- a/bin/express-cli.js
+++ b/bin/express-cli.js
@@ -182,9 +182,9 @@ function createApplication (name, dir) {
   }
 
   mkdir(dir, 'public')
-  mkdir(dir, 'public/javascripts')
+  mkdir(dir, 'public/scripts')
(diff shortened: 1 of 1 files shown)


dougwilson:
Hi @gumbol and thank you for your contribution! It looks like this may be incomplete, as the change is not actually updated in any of the templates, so those pages will end up looking in the wrong URL.

But even though, though I'm not disagreeing with your change, there are many books, videos, and similar that rely on the current names of these directories. The reason to change does not seem compelling enough to need to redo all those items to update for the new naming, imo.

gumbol:
thanks @dougwilson for the feedback! it seems like it is not complete yet. ill gladly defer

PR #295, opened 2021-04-02
Validated index.html

ian-knight-uofa:
Updated no-view static index.html page to pass w3c validation.

### DIFF
--- a/templates/js/index.html
+++ b/templates/js/index.html
@@ -1,13 +1,14 @@
+<!DOCTYPE html>
 <html>
 
-<head>
-  <title>Express</title>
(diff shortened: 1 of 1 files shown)


dougwilson:
Thank you. Is the indentation changes needed to pass validation too? Can you link to how you validated it so we can keep it valid when we update it if the goal is to have it pass that validation?

ian-knight-uofa:
Thanks!
The indentation is purely a style/consistency thing; happy to leave that it as-is.
The main issue is the missing DOCTYPE declaration.

Validated against the W3C validator http://validator.w3.org/
CLI/Java version available here https://validator.github.io/validator/

PR #306, opened 2022-05-27
Use `const` instead of `var`

p4nu:
Just refactoring `var` to be `const`


p4nu:
Closing as duplicate of https://github.com/expressjs/generator/pull/280

Issue #308, opened 2022-05-28
Running tests locally fails


p4nu:
When developing this I get an error which can't fit in a window... This could be a codespaces error but I want to be sure. Does some of the dependencies use Python2 here?

![image](https://user-images.githubusercontent.com/28947061/170815111-5eddf1da-3cd9-42f0-8461-b2a27316c53d.png)

Issue #296, opened 2021-05-19
Sass Generator not working out of the box


sschleemilch:
Hi,

I noticed that on a fresh Arch Linux installation, the express generator for the `sass` css engine is not working anymore.
That's probably because of missing support on the `node-sass-middleware` which seems to be quite out-of-date and not working properly anymore.

Has anyone a suggestion how to port the behavior to have a working sass middleware again with the most recent node?
Anyway, here is my setup to reproduce it:

Environment:
- `node -v` -> `v16.1.0`
- `node-gyp -v` -> `v8.0.0`
- `npm -v`-> `7.13.0`
- `express --version` -> `4.16.1` 

Steps:
- `express -c sass test`
- `cd test`
- `npm install`

Error:

> ........... 
npm ERR! /home/sebastian/.node-gyp/16.1.0/include/node/v8-internal.h:452:38: error: ‘remove_cv_t’ is not a member of ‘std’; did you mean ‘remove_cv’?
npm ERR!   452 |             !std::is_same<Data, std::remove_cv_t<T>>::value>::Perform(data);
npm ERR!       |                                      ^~~~~~~~~~~
npm ERR!       |                                      remove_cv
npm ERR! /home/sebastian/.node-gyp/16.1.0/include/node/v8-internal.h:452:38: error: ‘remove_cv_t’ is not a member of ‘std’; did you mean ‘remove_cv’?
npm ERR!   452 |             !std::is_same<Data, std::remove_cv_t<T>>::value>::Perform(data);
npm ERR!       |                                      ^~~~~~~~~~~
npm ERR!       |                                      remove_cv
npm ERR! /home/sebastian/.node-gyp/16.1.0/include/node/v8-internal.h:452:50: error: template argument 2 is invalid
npm ERR!   452 |             !std::is_same<Data, std::remove_cv_t<T>>::value>::Perform(data);
npm ERR!       |                                                  ^
npm ERR! /home/sebastian/.node-gyp/16.1.0/include/node/v8-internal.h:452:63: error: ‘::Perform’ has not been declared
npm ERR!   452 |             !std::is_same<Data, std::remove_cv_t<T>>::value>::Perform(data);
npm ERR!       |                                                               ^~~~~~~
.................

Interesting is also that in the end it reports about versions of `node` and `node-gyp` but reports the wrong `node-gyp` version in my opinion:
> npm ERR! gyp ERR! node -v v16.1.0
npm ERR! gyp ERR! node-gyp -v v3.8.0

[...]

PR #309, opened 2022-05-28
(Chore) Update node-sass-middleware to v1.0.1

p4nu:
This closes issues #308 and #296


p4nu:
I'll update changes to be backwards compatible with older nodejs versions.

PR #314, opened 2023-02-21
Update vulnerable dependencies

smorimoto:


joeyguerra:
I wonder if re running the failing job would result in a successful run?

PR #320, opened 2024-05-31
Basic changes are made

whoami1432:
Here I added the use strict for .js files and changed the var into const then while sending response and redirect we return it.

### DIFF
--- a/templates/js/app.js.ejs
+++ b/templates/js/app.js.ejs
@@ -1,17 +1,19 @@
+'use strict';
+
 <% if (view) { -%>
-var createError = require('http-errors');
+const createError = require('http-errors');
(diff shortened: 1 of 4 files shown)


UlisesGascon:
Hi @whoami1432! Thanks for opening a PR and help us improve the generator :smile: 

We currently don't support ES6 syntax as the project is supporting Node@0.10 (see: https://github.com/expressjs/generator/blob/master/.github/workflows/ci.yml#L14).

I enabled the CI, so you can debug better the changes that you are suggesting :+1:

alexanderankin:
@UlisesGascon how about creating a new repo in this org called "generator-esm"?

PR #328, opened 2024-12-13
fix: update default view engine from jade to pug

RushilJalal:

### DIFF
--- a/bin/express-cli.js
+++ b/bin/express-cli.js
@@ -467,7 +467,7 @@ function main (options, done) {
     if (options.view === true) {
       warning('the default view engine will not be jade in future releases\n' +
         "use `--view=jade' or `--help' for additional options")
-      options.view = 'jade'
+      options.view = 'pug'
     }
 
     // Generate application
--- a/test/cmd.js
+++ b/test/cmd.js
@@ -75,7 +75,9 @@ describe('express(1)', function () {
         '    "debug": "~2.6.9",\n' +
         '    "express": "~4.17.1",\n' +
         '    "http-errors": "~1.7.2",\n' +
-        '    "jade": "~1.11.0",\n' +
+        // replace default jade with pug
+        // '    "jade": "~1.11.0",\n' +
+        '    "pug": "~3.0.3",\n' +
         '    "morgan": "~1.10.0"\n' +
         '  }\n' +
         '}\n')


RushilJalal:
@UlisesGascon @IamLizu The checks are failing. I require some assistance on how to fix this...

UlisesGascon:
From what I can see:  
Screenshot from 2024-12-21 15-26-41

It seems like `pug@3.x` has dropped compatibility with the old versions of Node.js that we support (starting from 0.8):  
https://github.com/expressjs/generator/actions/runs/12314415858/job/34729590041?pr=328  

`pug` now only supports Node.js@9 and above ([see release notes](https://github.com/pugjs/pug/releases/tag/pug%403.0.0)). Ideally, we can include this change in the next major release or upgrade to a version of `pug` that remains compatible with the Node.js versions we currently support.

RushilJalal:
So what is the next step? Do I need to make any changes?

bin/express-cli.js
```diff
@@ -467,7 +467,7 @@ function main (options, done) {
     if (options.view === true) {
       warning('the default view engine will not be jade in future releases\n' +
```
IamLizu, line 468: Unnecessary warning if `jage` is no longer the default.

test/cmd.js
```diff
@@ -75,7 +75,9 @@ describe('express(1)', function () {
         '    "debug": "~2.6.9",\n' +
         '    "express": "~4.17.1",\n' +
         '    "http-errors": "~1.7.2",\n' +
-        '    "jade": "~1.11.0",\n' +
+        // replace default jade with pug
+        // '    "jade": "~1.11.0",\n' +
+        '    "pug": "~3.0.3",\n' +
```
IamLizu, line 80: Incompatible version of `pug`

IamLizu (changes requested):
>  we can include this change in the next major release or upgrade to a version of pug that remains compatible with the Node.js versions we currently support.

I think you should do the latter.

bin/express-cli.js
RushilJalal: Should I remove this warning completely?

test/cmd.js
RushilJalal: Why is the latest version of `pug` incompatible? Which version should it be instead?

PR #336, opened 2025-05-17
ci: apply OSSF Scorecard security best practices 

UlisesGascon:
related https://github.com/expressjs/security-wg/issues/2

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: monthly
+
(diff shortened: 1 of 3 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: monthly
+
+  - package-ecosystem: npm
```
bjohansebas, line 8: We need to make Dependabot stop opening pull requests that update to major versions until we resolve the linter issue

UlisesGascon: Patched in f500009. Thanks for the ping!

